PR #1, opened 2023-01-11
Boilerplate

paramite:
Adds boilerplate code according to [1].

[1] https://github.com/openstack-k8s-operators/docs/blob/main/new_operator.md


jlarriba (changes requested):
It works perfectly with the operator development cycle. 

However, I dont think that we really want two apis and two controllers (ceilometernotificationagent and ceilometercentralagent), as I favor much more the deployment of both agents in a single Pod, with a single API and single controller. Can we just have one "ceilometer" API and a single ceilometer_controller.go?

Good work!

jlarriba (approved):

PR #15, opened 2023-02-06
Drop unused template for transport_url

kajinamit:
The transport_url options are set by init.sh from secrets and we do not use template mechanism for the options.

Signed-off-by: Takashi Kajinami <tkajinam@redhat.com>

### DIFF
--- a/templates/ceilometer/config/ceilometer.conf
+++ b/templates/ceilometer/config/ceilometer.conf
@@ -4,7 +4,6 @@ tenant_name_discovery=False
 debug=True
 log_dir=/var/log/ceilometer
 rpc_response_timeout=60
-transport_url={{ .transportURL }}
 polling_namespaces=central
 
 [service_credentials]
@@ -36,7 +35,6 @@ notify_address_prefix=
 
 [oslo_messaging_notifications]
 driver=messagingv2
-transport_url={{ .transportURL }}
 topics=notifications
 
 [publisher]


jlarriba (commented):
It is true that this is not used, as it is set up on init.sh, but I would like to leave this here as a reference to remember that this variable is being set in init.sh, it is not just missing.

The Glance operator does the same and I like it for the sake of completeness.

kajinamit:
Hmm. I was looking at the other operators but I could not find the same usage (even in Glance).

>  as a reference to remember that this variable is being set in init.sh
IMO the current placeholder is misleading because it looks like there should be the variable to override the value by template. Also it can leave invalid config value in case config override does not work.

We may probably use comments instead of putting real lines. Does it make sense to you ?

jlarriba (commented):
In Glance there are multiples examples on the glance-api.conf file, like this one: 

https://github.com/openstack-k8s-operators/glance-operator/blob/master/templates/glance/config/glance-api.conf#L21

kajinamit:
> In Glance there are multiples examples on the glance-api.conf file, like this one:
> 
> https://github.com/openstack-k8s-operators/glance-operator/blob/master/templates/glance/config/glance-api.conf#L21

That is replaced by the template function when generating config hash. https://github.com/openstack-k8s-operators/glance-operator/blob/master/controllers/glanceapi_controller.go#L512
The current glance-api.conf does not include entries replaced by init.sh.

jlarriba (approved):
Ohhh, you are completely right sir. I was misled and forgot about the template parameters.

Thank you for your kind guidance on this issue.

PR #46, opened 2023-03-07
Use constant to define image pull policy

kajinamit:
... instead hard-coding the string in our own logic.

### DIFF
--- a/pkg/ceilometer/deployment.go
+++ b/pkg/ceilometer/deployment.go
@@ -81,7 +81,7 @@ func Deployment(
 	var replicas int32 = 1
 
 	centralAgentContainer := corev1.Container{
-		ImagePullPolicy: "Always",
+		ImagePullPolicy: corev1.PullAlways,
 		Command: []string{
 			"/bin/bash",
 		},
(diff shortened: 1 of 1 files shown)


pkg/ceilometer/deployment.go
```diff
@@ -81,7 +81,7 @@ func Deployment(
 	var replicas int32 = 1
 
 	centralAgentContainer := corev1.Container{
-		ImagePullPolicy: "Always",
+		ImagePullPolicy: corev1.PullAlways,
```
kajinamit, line 84: by the way, why do we need this in the first place ? can't we use the existing image ?

paramite: You mean image passed via `instance` spec?

jlarriba: We want the operator to always download the image, mostly is utterly useful during development.

jlarriba (approved):

pkg/ceilometer/deployment.go
kajinamit: The current hard-code makes worker nodes to pull ceilometer images always, which is quite insufficient. (eg. a container should be restarted after config change )

I'm fine with leaving this for testing purpose, but we should properly keep TODO so that we remove this before GA.
# though I'm not too sure how much it is useful for us, because ceilometer containers are built by a different pipeline

PR #64, opened 2023-03-29
Moved ceilometer-operator to telemetry-operator

jlarriba:


paramite:
So I won't be the one adding a green flag to this. It's just a name and a fact that we will have sg-core and aodh managed by the operator is not worth the change. Renaming brings number of changes necessary not only in this repo, just for the sake of name. Prometheus-operator also manages not only prometheus, but alertmanager, thanos-ruler, etc. ... my two cents.

PR #74, opened 2023-04-13
remove duplicates from Dockerfile

yazug:
remove duplicate FROM and RUN ... go build lines

This should bring us closer to inline with what other operators have currently.


jlarriba (approved):

abays (approved):
/lgtm

PR #79, opened 2023-04-18
Replace TripleO images by TCIB images

kajinamit:
We imported container build tools from TripleO as TCIB[1]. Now we are retiring TripleO and should complete the migration.

[1] https://github.com/openstack-k8s-operators/tcib

### DIFF
--- a/api/bases/telemetry.openstack.org_telemetries.yaml
+++ b/api/bases/telemetry.openstack.org_telemetries.yaml
@@ -36,7 +36,7 @@ spec:
             description: CeilometerSpec defines the desired state of Ceilometer
             properties:
               centralImage:
-                default: quay.io/tripleomastercentos9/openstack-ceilometer-central:current-tripleo
+                default: quay.io/podified-antelope-centos9/openstack-ceilometer-central:current-podified
                 type: string
               customServiceConfig:
                 default: '# add your customization here'
@@ -58,7 +58,7 @@ spec:
                 default: A ceilometer agent
                 type: string
               initImage:
-                default: quay.io/tripleomastercentos9/openstack-ceilometer-central:current-tripleo
+                default: quay.io/podified-antelope-centos9/openstack-ceilometer-central:current-podified
                 type: string
               networkAttachmentDefinitions:
                 description: NetworkAttachmentDefinitions list of network attachment
@@ -67,7 +67,7 @@ spec:
                   type: string
                 type: array
               notificationImage:
-                default: quay.io/tripleomastercentos9/openstack-ceilometer-notification:current-tripleo
+                default: quay.io/podified-antelope-centos9/openstack-ceilometer-notification:current-podified
                 type: string
               passwordSelector:
                 default:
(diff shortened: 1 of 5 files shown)


api/bases/telemetry.openstack.org_telemetries.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             description: CeilometerSpec defines the desired state of Ceilometer
             properties:
               centralImage:
```
kajinamit, line 38: Assuming we onboard more services(such as aodh) to telemetry operator, should we rename this to ceilometerCentralImage ?

api/bases/telemetry.openstack.org_telemetries.yaml
```diff
@@ -36,7 +36,7 @@ spec:
...
-                default: quay.io/tripleomastercentos9/openstack-ceilometer-central:current-tripleo
+                default: quay.io/podified-antelope-centos9/openstack-ceilometer-central:current-podified
                 type: string
               customServiceConfig:
```
kajinamit, line 41: Probably we need the prefix for the other options like this. (ceilometerCustomServiceConfig)

api/bases/telemetry.openstack.org_telemetries.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             description: CeilometerSpec defines the desired state of Ceilometer
```
kajinamit, line 36: Should we update this ? (TelemetrySpec ?)

kajinamit (commented):
Just leaving a few off-topic items I noticed.

api/bases/telemetry.openstack.org_telemetries.yaml
paramite: Having CeilometerSpec enables us to add AodhSpec  and have CeilometerCentral separately from Aodh. telemetry-operator won't be managing just one type of service.

api/bases/telemetry.openstack.org_telemetries.yaml
paramite: Not necessarily. If we keep CeilometerCentral and Aodh separated.

api/bases/telemetry.openstack.org_telemetries.yaml
paramite: ditto

paramite (approved):

api/bases/telemetry.openstack.org_telemetries.yaml
kajinamit: My point is that this is the spec schema for kind:Telemetry right ? Then this should not be specific to Ceilometer.

api/bases/telemetry.openstack.org_telemetries.yaml
kajinamit: This is the top level key of Kind: Telemetry so we can't really separate aodh and ceilometer, iiuc.

paramite: Yes you're correct. Replied in haste.

PR #82, opened 2023-05-02
Webhooks for default images

abays:
Default container images are now handled via env vars and internal webhooks, as per the pattern we are using for all NextGen operators.  Also attempts to modernize and clean-up CI workflows as per that pattern in other operators.


.github/workflows/pre-commit.yaml
abays, line 1: This is now handled in prow

jlarriba (approved):
/lgtm

PR #88, opened 2023-05-10
Remove kubebuilder CeilometerCompute default images

abays:
Removes some `kubebuilder` defaults that were inadvertently added to `CeilometerCompute` after the webhooks patch.  Also fixes `make kustomize` so it can re-run without manually removing `kustomize` (a pattern we use in other operators).


leifmadsen (approved):
+1 (deferred +2 approval to cores)

paramite (approved):

PR #105, opened 2023-06-01
Webhook cleanup

abays:
Cleaning up webhooks to conform to the finalized pattern used across all operators.  If/when we refactor the CRDs to account for control and data plane division, we can adjust the webhooks again (but still use this finalized pattern).


jlarriba (approved):
/lgtm

PR #106, opened 2023-06-01
Every CRD can be ran independently

jlarriba:
A Telemetry object is not required anymore to deploy any of the CRs. It is now only an optional umbrella CR that can be used or not.


jlarriba:
Im going to merge https://github.com/openstack-k8s-operators/telemetry-operator/pull/105 and create three new webhook objects, one per CRD, and remove the telemetry webhook

api/v1beta1/telemetry_webhook.go
abays, line 1: It might be a good idea to leave the `Telemetry` CRD webhook as-is if we're not going to remove the CRD itself.  That way if someone does create one, the defaults will be visibly populated within it (i.e.`oc get telemetry <name> -o yaml`) as opposed to only in the sub-resources (i.e. `CeilometerCompute`) because the sub-resource webhooks caught and handled them later on.  If the `Telemetry` webhook remains, it's not a problem either if it fires and sets the defaults and then later the sub-resource webhooks fires and process some of the same defaulting again.  It should all still work properly.

Here are examples [1] [2] [3] [4] [5] from Nova, which has webhooks at both the top-level `Nova` CRD and for the sub-resources.

[1] https://github.com/openstack-k8s-operators/nova-operator/blob/ebaeca670b17a751bfb01a386f4e1f4811588dc5/api/v1beta1/common_types.go#L26-L36
[2] https://github.com/openstack-k8s-operators/nova-operator/blob/ebaeca670b17a751bfb01a386f4e1f4811588dc5/api/v1beta1/common_types.go#L154-L218
[3] https://github.com/openstack-k8s-operators/nova-operator/blob/ebaeca670b17a751bfb01a386f4e1f4811588dc5/api/v1beta1/nova_webhook.go#L32-L41
[4] https://github.com/openstack-k8s-operators/nova-operator/blob/ebaeca670b17a751bfb01a386f4e1f4811588dc5/api/v1beta1/novaapi_webhook.go#L32-L37
[5] https://github.com/openstack-k8s-operators/nova-operator/blob/ebaeca670b17a751bfb01a386f4e1f4811588dc5/main.go#L134-L135

jlarriba:
The golangci error does not seem related to my code.

@Sandeepyadav93 is there something failing on the CI?

jlarriba:
At last it worked fine, all test passed

api/v1beta1/ceilometercentral_types.go
```diff
@@ -120,3 +138,30 @@ func (instance CeilometerCentral) IsReady() bool {
...
+}
+
+// SetupDefaultsCeilometerCentral - initializes any CRD field defaults based on environment variables (the defaulting mechanism itself is implemented via webhooks)
+func SetupDefaultsCeilometerCentral() {
+	// Acquire environmental defaults and initialize Telemetry defaults with them
+	ceilometercentralDefaults := CeilometerCentralDefaults{
+		CentralContainerImageURL:      util.GetEnvVar("CEILOMETER_CENTRAL_IMAGE_URL_DEFAULT", CeilometerCentralContainerImage),
+		CentralInitContainerImageURL:  util.GetEnvVar("CEILOMETER_CENTRAL_INIT_IMAGE_URL_DEFAULT", CeilometerCentralInitContainerImage),
+		SgCoreContainerImageURL:       util.GetEnvVar("CEILOMETER_SGCORE_IMAGE_URL_DEFAULT", CeilometerSgCoreContainerImage),
+	}
```
abays, line 164: Don't we need `Notification` image defaults here as well?

jlarriba:
@abays To answer your question, I have already been working on the modifications I plan to introduce on openstack-operator. It is indeed basically replacing a Telemetry deployment with a CeilometerCentral deployment, but calling it just Ceilometer, as from the openstack-operator perspective it does not need to know there are more.

I just submitted my changes to a draft PR to show the code: https://github.com/openstack-k8s-operators/openstack-operator/pull/339

Regarding RBAC,I am going to look at that. I did not introduced the capacity for the user to pass a ServiceAccount and I had to modify it during separation. I tested that everything worked correctly, but I will review because it might only be working by chance (just because the ServiceAccount name specified in the CR is actually telemetry-<CeilometerCentral instance name>). So I will fix it.

To address the second concern about RBAC: yes the idea of the whole telemetry system using just one ServiceAccount is desired. I dont really see why they should be using separate ServiceAccounts nor do I see any advantages on getting the different CRs use different SAs. 

Passing the SA to the ansibleEE execution is needed because ansibleEE init container needs to do merging in files that some random user does not have permissions to do. In this case we need a SA with anyuid permissions, and the telemetry SA already has that. As the ansibleEE runner pod can be executed by very different operators, it seems reasonable for them to run that runner with their own SA if that is their desire.

abays:
> @abays To answer your question, I have already been working on the modifications I plan to introduce on openstack-operator. It is indeed basically replacing a Telemetry deployment with a CeilometerCentral deployment, but calling it just Ceilometer, as from the openstack-operator perspective it does not need to know there are more.
> 
> I just submitted my changes to a draft PR to show the code: [openstack-k8s-operators/openstack-operator#339](https://github.com/openstack-k8s-operators/openstack-operator/pull/339)
> 
> Regarding RBAC,I am going to look at that. I did not introduced the capacity for the user to pass a ServiceAccount and I had to modify it during separation. I tested that everything worked correctly, but I will review because it might only be working by chance (just because the ServiceAccount name specified in the CR is actually telemetry-). So I will fix it.
> 
> To address the second concern about RBAC: yes the idea of the whole telemetry system using just one ServiceAccount is desired. I dont really see why they should be using separate ServiceAccounts nor do I see any advantages on getting the different CRs use different SAs.
> 
> Passing the SA to the ansibleEE execution is needed because ansibleEE init container needs to do merging in files that some random user does not have permissions to do. In this case we need a SA with anyuid permissions, and the telemetry SA already has that. As the ansibleEE runner pod can be executed by very different operators, it seems reasonable for them to run that runner with their own SA if that is their desire.

Ack, sounds good!  Thanks for all the explanations.

jlarriba:
Regarding RBAC, I have realized that everything was working because everything was using the "old" SA telemetry-operator-telemetry in my environment that was never removed, and nothing was using the SA created by the controllers. Each controller creates its own SA, as the user is able to specify it now on their Specs instead of the common Telemetry spec, so we cannot control whether it is the same or not.

This creates a difficult situation, because I liked the idea of everything under the telemetry umbrella to use the same SA, but giving the user the opportunity to decide, this is not always to be like this.

Allowing the user define the SA for each CRD is very important? This was introduced by a PR from @dprince

config/webhook/manifests.yaml
```diff
@@ -5,6 +5,66 @@ metadata:
...
+    resources:
+    - ceilometercentrals
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-telemetry-openstack-org-v1beta1-ceilometercentral
```
abays, line 34: `s/ceilometercentral/ceilometercompute`

SeanMooney (changes requested):
can you incoperate the removal of the defaults for DataplaneSSHSecret and DataplaneInventoryConfigMap from 
https://github.com/openstack-k8s-operators/telemetry-operator/pull/119/files

jlarriba:
@SeanMooney Sure, removed the defaults

Issue #117, opened 2023-06-09
The telemetry operator should not deploy the comptue compoent unless asked.


SeanMooney:
as of https://github.com/openstack-k8s-operators/telemetry-operator/commit/6ee1b2f73696d314834bd5b9413d9e20745f56cd the telemetry operator is now deploying the infra-compute
components even when there are no compute nodes.

https://github.com/openstack-k8s-operators/telemetry-operator/blob/b2d4fc55966ad66e3fdab94f8345e8698cd69979/controllers/telemetry_controller.go#L375-L386

While the exact mechanism to control this, in general, is still under active discussion today the agreed acrchitreure is that if a service operator need to run Ansible against a dataplane node it would only do so if the dataplane operator create  a CR that the service operator reconciles.

while we are discussion selvel options the one that is likely to replace that is tagging OpenstackDataplaneNodes or OpenstackDataplaneRoles with an annotation that would cause the service operator to take the approate action

i.e. openstack.telemetry.infraCompute=true would be applied to the dataplane nodes that should have the infra compute ansible applied.

SeanMooney:
the same issue extis for ceilometier comptue

https://github.com/openstack-k8s-operators/telemetry-operator/blob/b2d4fc55966ad66e3fdab94f8345e8698cd69979/controllers/telemetry_controller.go#L375-L386

however it does not create an AnsibleEE by default



the relevant section form the OpenStack control plane is before the defaulting webhooks ran was

telemetry:
    template:
      ceilometerCentral:
        passwordSelector:
          service: CeilometerPassword
        secret: osp-secret
        serviceUser: ceilometer

and this became after.


  telemetry:
    enabled: true
    template:
      ceilometerCentral:
        centralImage: quay.io/podified-antelope-centos9/openstack-ceilometer-central:current-podified
        customServiceConfig: '# add your customization here'
        description: A ceilometer agent
        initImage: quay.io/podified-antelope-centos9/openstack-ceilometer-central:current-podified
        notificationImage: quay.io/podified-antelope-centos9/openstack-ceilometer-notification:current-podified
        passwordSelector:
          service: CeilometerPassword
        secret: osp-secret
        serviceAccount: ""
        serviceUser: ceilometer
        sgCoreImage: quay.io/infrawatch/sg-core:latest
      ceilometerCompute:
        computeImage: quay.io/podified-antelope-centos9/openstack-ceilometer-compute:current-podified
[...]

PR #119, opened 2023-06-09
adjust CRDs to update defaults

SeanMooney:
This change makes the CeilometerCompute and InfraCompute
filed optional and pointers in the TelemetrySepc and then modifes
the telemetry contoler to only reconsile them if they are != nil

second this change removes the defalt values for the
DataplaneSSHSecret and DataplaneInventoryConfigMap fields
in the CeilometerComputeSpec and InfraComputeSpec and updates
the input validation of the contolers to ensure they are presenset
if a value is provided as part of the input validation status.

Related: #117


SeanMooney:
i have not actully tested this by the way but something like that would resolve the main defaulting issues in #117 i think

jlarriba:
The objective of this PR is the same as the one that generated https://github.com/openstack-k8s-operators/telemetry-operator/pull/106. In either case, it allows the openstack-operator to only deploy CeilometerCentral without having to also carry InfraCompute and CeilometerCompute, which should be deployed by the dataplane operator. In this case, we will continue by merging the other one.

SeanMooney:
https://github.com/openstack-k8s-operators/telemetry-operator/pull/106 now has the default changes for the inventory and ssh key so ill close this

PR #122, opened 2023-06-13
Remove playbooks

jlarriba:


controllers/ceilometercompute_controller.go
```diff
@@ -486,6 +452,25 @@ func (r *CeilometerComputeReconciler) transportURLCreateOrUpdate(instance *telem
...
+	data["extravars"] = fmt.Sprintf("deploy_target_host: %s", "all")
+
+	configMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-ceilometer-extravars", telemetry.ServiceName),
+			Namespace: instance.Namespace,
+		},
+		Data: data,
+	}
+	_, err := controllerutil.CreateOrPatch(ctx, h.GetClient(), configMap, func() error { return nil })
```
abays, line 466: I think the recommended pattern for `CreateOrPatch` is to create the base object (object meta name/namespace) outside the func, then set specifics (in this case, `Data`) inside the func.  We might want to do that here unless there's a specific reason to not follow the pattern.

controllers/infracompute_controller.go
```diff
@@ -171,6 +168,17 @@ func (r *InfraComputeReconciler) reconcileDelete(ctx context.Context, instance *
 		return ctrl.Result{}, err
 	}
 
+	// Delete the extravars cm when the infracompute is deleted
+	extravars := &corev1.ConfigMap{}
+	err = r.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s-infra-extravars", telemetry.ServiceName), Namespace: instance.Namespace}, extravars)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
abays, line 176: Same thing here with regards to checking "not found" `err` and allowing it to pass if so

controllers/ceilometercompute_controller.go
jlarriba: I dont think I understand this fully. The configmap lib-common library does not have a "NewConfigMap" function, as it has for example, on Deployments so to implement the pattern you are saying, I would need to create a NewConfigMap func myself to just create the ObjectMeta, which does not make a lot of sense. 

As the configmap library is only providing an EnsureConfigMap function that include template processing capabilities, I am just doing the same Nova is doing for creating ConfigMaps that do not require templating capabilities, as in https://github.com/openstack-k8s-operators/nova-operator/blob/7fdafd5549c4a02eed7ca6e35067256b5e4d754e/controllers/novaexternalcompute_controller.go#L526

jlarriba: I think I understood what you were saying. Pushed the corresponding change.

controllers/ceilometercompute_controller.go
```diff
@@ -486,6 +456,26 @@ func (r *CeilometerComputeReconciler) transportURLCreateOrUpdate(instance *telem
...
+	configMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-ceilometer-extravars", telemetry.ServiceName),
+			Namespace: instance.Namespace,
+		},
+		Data: map[string]string{},
+	}
+
+	_, err := controllerutil.CreateOrPatch(ctx, h.GetClient(), configMap, func() error {
+		configMap.Data["extravars"] = fmt.Sprintf("deploy_target_host: %s", "all")
```
abays, line 469: Just one last thing.  Could just remove the `fmt.Sprintf` and use a pre-formed string since all the pieces here are static.

jlarriba: That makes a lot of sense. Done.

abays (approved):
/lgtm

PR #142, opened 2023-07-10
Remove unused owner reference

kajinamit:
Ceilometer central controller does not create KeystoneAPI CR but it only read its status. Also, the controller does not create own Secret but just looks up the existing global secret.


kajinamit:
This will be addressed by https://github.com/openstack-k8s-operators/telemetry-operator/pull/142

PR #159, opened 2023-08-02
Update some spec fields for consistency

kajinamit:
In most of our operators the `spec.secret` field is required. Let's follow that instead of adding the default value.

Also currently `spec.transportURLSecret` is updated by the controller but we should not update Spec but should use Status. This fixes it and ensure the value is stored in the new status field.

Finally This removes `spec.serviceAccount` and `spec.Description` which are not used.


jlarriba (approved):
I have tested and everything continues to be working correctly

PR #167, opened 2023-08-31
Removing references to ansibleee Role struct

jpodivin:
The services will now be deployed using playbooks in edpm-ansible collection.


slagle (commented):
:+1:

yadneshk (approved):

PR #168, opened 2023-08-31
Restoring code for reconcilliation on hash change

jpodivin:
I'm not sure why was this removed in the first place, but it has been commented out since the file was added. 

Depending on the reason we may want to remove it entirely, replace it, or restore it like here.


controllers/ceilometercentral_controller.go
```diff
@@ -356,18 +356,14 @@ func (r *CeilometerCentralReconciler) reconcileNormal(ctx context.Context, insta
 	// and a restart/recreate is required.
 	//
 	inputHash, hashChanged, err := r.createHashOfInputHashes(ctx, instance, configMapVars)
-	fmt.Printf("hashChanged: %v\n", hashChanged)
```
jlarriba, line 359: Having this line here is an error, its just a dev debug line that must be removed.

PR #176, opened 2023-09-07
Renamed the playbook in edpm_ansible

jlarriba:


jlarriba:
This is a parallel PR to avoid https://github.com/openstack-k8s-operators/edpm-ansible/pull/312 to break things here. It is temporary, as the infracompute and the ceilometercompute APIs will disappear as the compute management will be delegated to dataplane-operator.

jpodivin (approved):

paramite (approved):

PR #193, opened 2023-09-22
Fix the sg-core version, using latest is no sane

jlarriba:


abays (approved):
Thanks, this has been causing problems in CI

/lgtm

PR #194, opened 2023-09-22
Add aodh to autoscaling

vyzigold:
This PR adds aodh to the autoscaling api.

Because the autoscaling controller got quite big, I decided to split it to 3 (4 in the future) files. controllers/autoscaling_controller.go for some common code and /controllers/prometheus_controller.go, /controllers/aodh_controller for the prometheus and aodh related code.

The controller passes the prometheus host and port values to the aodh-evaluator container as expected by the observability client, which will be used for autoscaling in the future.

The functionality of aodh is a bit hard to test at the moment, because we don't have a supported storage for it (gnocchi at this moment, prometheus is coming in the future) we can't create any alarms. But I was able to at least execute `openstack alarm list` from the openstackclient container. This resulted in an empty response (expected), but most importantly it didn't produce any error in any of the aodh containers or the operator pod.

There is also a related patch to install_yamls to add the required aodh and database passwords to the osp-secret: https://github.com/openstack-k8s-operators/install_yamls/pull/541


vyzigold:
I'll close and reopen this exact PR to get it tested by the new functional tests.

PR #201, opened 2023-10-06
Use database helpers from mariadb-operator/api

vyzigold:
The database helpers got moved from lib-common to mariadb-operator to deal with circular dependencies: https://github.com/openstack-k8s-operators/lib-common/pull/352

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,8 +13,7 @@ require (
 	github.com/openstack-k8s-operators/infra-operator/apis v0.1.1-0.20230905074428-c6aefc16dd01
 	github.com/openstack-k8s-operators/keystone-operator/api v0.1.1-0.20230830083045-d73d07cca617
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.1.1-0.20230927082538-4f614f333d17
-	github.com/openstack-k8s-operators/lib-common/modules/database v0.1.0
-	github.com/openstack-k8s-operators/mariadb-operator/api v0.1.0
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -13,8 +13,7 @@ require (
 	github.com/openstack-k8s-operators/infra-operator/apis v0.1.1-0.20230905074428-c6aefc16dd01
 	github.com/openstack-k8s-operators/keystone-operator/api v0.1.1-0.20230830083045-d73d07cca617
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.1.1-0.20230927082538-4f614f333d17
-	github.com/openstack-k8s-operators/lib-common/modules/database v0.1.0
-	github.com/openstack-k8s-operators/mariadb-operator/api v0.1.0
+	github.com/openstack-k8s-operators/mariadb-operator/api v0.3.0
```
jlarriba, line 16: We should not be fixing into a version, use the latest v0.1.1-0 as with the rest of the operators, like it is being done here: https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/go.mod#L10

jlarriba (changes requested):

go.mod
jlarriba: yes, this is the behavior we are seeing. I think we could just go with the latest tag as that's default behaviour for go mod.

gibizer: Yeah, I think this is OK to land. Renovate will bump over 0.3.0 with a pseudoversion when there is a newer commit

jlarriba (approved):

PR #202, opened 2023-10-11
Don't label all host files with container_file_t:

rabi:
Resolves: https://issues.redhat.com/browse/OSP-29586

### DIFF
--- a/templates/ceilometercompute/config/ceilometer_agent_compute.json
+++ b/templates/ceilometercompute/config/ceilometer_agent_compute.json
@@ -13,13 +13,13 @@
         "/var/lib/openstack/config/ceilometer/ceilometer-agent-compute.json:/var/lib/kolla/config_files/config.json:z",
         "/var/log/containers/ceilometer:/var/log/containers/ceilometer:z",
         "/run/libvirt:/run/libvirt:shared,z",
-        "/etc/hosts:/etc/hosts:z",
-        "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:z",
-        "/etc/localtime:/etc/localtime:z",
-        "/etc/pki/ca-trust/extracted:/etc/pki/ca-trust/extracted:z",
-        "/etc/pki/ca-trust/source/anchors:/etc/pki/ca-trust/source/anchors:z",
-        "/dev/log:/dev/log:z",
-        "/etc/pki/tls/cert.pem:/etc/pki/tls/cert.pem:z",
-        "/etc/pki/tls/certs/ca-bundle.crt:/etc/pki/tls/certs/ca-bundle.crt:z"
+        "/etc/hosts:/etc/hosts:ro",
+        "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:ro",
+        "/etc/localtime:/etc/localtime:ro",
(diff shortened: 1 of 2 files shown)


templates/ceilometercompute/config/ceilometer_agent_compute.json
```diff
@@ -15,7 +15,7 @@
         "/run/libvirt:/run/libvirt:shared,z",
         "/etc/hosts:/etc/hosts:z",
         "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:z",
-        "/etc/localtime:/etc/localtime:z",
+        "/etc/localtime:/etc/localtime:ro",
```
gibizer, line 18: will the other `:z ` in these files cause the same issue?

rabi: Probably not. I changed a few more based on what was in TripleO.

templates/ceilometercompute/config/ceilometer_agent_ipmi.json
```diff
@@ -13,13 +13,13 @@
...
-        "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:z",
-        "/etc/localtime:/etc/localtime:z",
-        "/etc/pki/ca-trust/extracted:/etc/pki/ca-trust/extracted:z",
-        "/etc/pki/ca-trust/source/anchors:/etc/pki/ca-trust/source/anchors:z",
-        "/dev/log:/dev/log:z",
-        "/etc/pki/tls/cert.pem:/etc/pki/tls/cert.pem:z",
-        "/etc/pki/tls/certs/ca-bundle.crt:/etc/pki/tls/certs/ca-bundle.crt:z"
+        "/etc/hosts:/etc/hosts:ro",
+        "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:ro",
+        "/etc/localtime::ro",
```
jlarriba, line 18: Shouldnt this be /etc/localtime:/etc/localtime:ro?

jlarriba (changes requested):

templates/ceilometercompute/config/ceilometer_agent_ipmi.json
rabi: sorry, not sure how that disappeared.

templates/ceilometercompute/config/ceilometer_agent_compute.json
gibizer: thanks.

gibizer (approved):
looks good to me.

templates/ceilometercompute/config/ceilometer_agent_compute.json
jlarriba: I have had a lot of troubles with podman and not using either :z or :Z to set up selinux permissions of the files. When I was testing this, the podman container will not mount any files that were not annotated with either :z or :Z, so are we sure this will work?

Yesterday I was testing a new container in a compute node and the only way I could get the container to mount the yaml configuration file was annotating the mount with :z, with :ro the mount did not appear on the container.

rabi: I've not seen that behavior, we use `:ro` with podman since eternity :smile:

[1] https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/containers-common.yaml#L124

jlarriba (approved):

PR #212, opened 2023-10-26
Make kuttl suite compatible with install_yamls

paramite:
Prow CI job executing KUTTL test suites has different structure and logic then what we have. This patch is making it work in both cases.


paramite:
recheck

PR #230, opened 2023-11-06
Set default image url for node exporter image

yadneshk:
Without this change when telemetry services are initialized with default values, "nodeExporterImage" holds no value.


jlarriba:
This value is not used anymore. This property was used to pass it to ansibleee in the compute nodes back when the controller was executing ansibleee by itself. This is not currently the case and that execution was delegated to the dataplane-operator which does not pass CR variables to ansible execution. This image is defined as an ansible var in edpm_ansible [1]. 

What we might do is remove the node_exporter image property from Spec alltogether. It is still there because I have hope in that dataplane-operator will eventually enable a way to pass CR variables to ansible execution. See [2] for related work on this regard.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/defaults/main.yml#L24 
[2] https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/246

jlarriba:
Closing this PR, superseded by the change described in https://issues.redhat.com/browse/OSPRH-1804

PR #232, opened 2023-11-10
KUTTL tests for Ceilometer and Telemetry API

paramite:
Adds remaining KUTTL suites and sets default suite to test Telemetry CR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -332,7 +332,6 @@ kuttl-install:
 	export PATH="$${KREW_ROOT:-$$HOME/.krew}/bin:$$PATH" && kubectl krew install kuttl
 	echo "Place 'export PATH=$${KREW_ROOT:-$$HOME/.krew}/bin:$$PATH' to your ~/.bashrc"
 
-# TODO(mmagr): change this to default once there is default suite
 KUTTL_SUITE ?= default
 KUTTL_NAMESPACE ?= telemetry-kuttl-$(KUTTL_SUITE)
 KUTTL_SUITE_DIR ?= tests/kuttl/suites/$(KUTTL_SUITE)
--- a/Makefile
+++ b/Makefile
@@ -333,7 +333,7 @@ kuttl-install:
 	echo "Place 'export PATH=$${KREW_ROOT:-$$HOME/.krew}/bin:$$PATH' to your ~/.bashrc"
 
 # TODO(mmagr): change this to default once there is default suite
-KUTTL_SUITE ?= autoscaling
+KUTTL_SUITE ?= default
 KUTTL_NAMESPACE ?= telemetry-kuttl-$(KUTTL_SUITE)
 KUTTL_SUITE_DIR ?= tests/kuttl/suites/$(KUTTL_SUITE)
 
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -333,7 +333,7 @@ kuttl-install:
 	echo "Place 'export PATH=$${KREW_ROOT:-$$HOME/.krew}/bin:$$PATH' to your ~/.bashrc"
 
 # TODO(mmagr): change this to default once there is default suite
```
vyzigold, line 335: I guess this can be deleted now?

tests/kuttl/suites/default/deps/rhobs.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
```
vyzigold, line 2: Maybe I should note, that in the future, I think we might want to use the OBO from the community-operators instead of using a custom catalog source. I tried to use it while developing the autoscaling without success and it seems there is still the same version, which didn't seem to work for me back then (0.0.25) while in the OBO repo I see latest version 0.0.28. So there is nothing to do now, just something to keep in mind in the future.

vyzigold (commented):
Otherwise this LGTM, although I don't know much about kuttl

Makefile
paramite: true

tests/kuttl/suites/default/deps/rhobs.yaml
paramite: Yes indeed. This is just a workaround to make it work now.

tests/kuttl/suites/default/tests/01-deploy.yaml
```diff
@@ -0,0 +1,22 @@
...
+  autoscaling:
+    enabled: true
+    prometheus:
+      deployPrometheus: true
+    aodh:
+      secret: osp-secret
+      apiImage: "quay.io/mmagr/openstack-aodh-api:current-podified"
+      evaluatorImage: "quay.io/mmagr/openstack-aodh-evaluator:current-podified"
+      notifierImage: "quay.io/mmagr/openstack-aodh-notifier:current-podified"
+      listenerImage: "quay.io/mmagr/openstack-aodh-listener:current-podified"
```
abays, line 15: Is there another source besides a personal repo for these?

vyzigold: The feature in aodh needs to be backported from master to 2023.1 for us to be able to actually use it. But if we are just checking that deployments and services are created, then it shouldn't matter which images we use and we should indeed probably use the antelope images here.

paramite: In the time of writing the tests Aodh and Heat patches were still on review. These days images built with master branches of both should have the autoscaling bits in. Backporting to stable/antelope will be a bit problematic though, and maybe impossible. As far as [1] is saying we cannot backport new feature (Heat, Aodh) nor new dependency (python-observabilityclient) on upstream side. My thought was to create a TCIB fork with a GH actions workflow, which will build images and push to quay.io/infrawatch on regular basis. 

[1] https://docs.openstack.org/project-team-guide/stable-branches.html#active-maintenance (point 1.)

PR #240, opened 2023-11-20
Rewrite telemetry resource

vyzigold:
Rewrite the telemetry resource to use a similar pattern to the openstack control plane resource. The telemetry resource is an umbrela resource for other telemetry related resources, which for right now are "ceilometer" and "autoscaling". These resources can be enabled and customized by using the "enabled" and "template" fields similarly to how openstack services are enabled and customized in the openstack control plane CR.

The "secret" field in the telemetry resource can be used to set the "secret" fields for both ceilometer and autoscaling.

The autoscaling is disabled by default and ceilometer is enabled by default.

Look at the config/samples/telemetry_v1beta1_telemetry.yaml for an example CR.


api/v1beta1/telemetry_types.go
```diff
@@ -45,13 +45,46 @@ type PasswordsSelector struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	// Enabled - Whether OpenStack Ceilometer service should be deployed and managed
+	Enabled bool `json:"enabled"`
+
+	// +kubebuilder:validation:Optional
+	//+operator-sdk:csv:customresourcedefinitions:type=spec
+	// Template - Overrides to use when creating the OpenStack Ceilometer service
+	Template CeilometerSpec `json:"template,omitempty"`
```
jlarriba, line 73: To avoid having to redeclare "template:", this should be an inline attribute:
```
CeilometerSpec `json:",omitempty"`
```

In the same way as, for example, Cinder is doing:

https://github.com/openstack-k8s-operators/cinder-operator/blob/main/api/v1beta1/cinderapi_types.go#L54

config/samples/telemetry_v1beta1_telemetry.yaml
```diff
@@ -2,8 +2,21 @@ apiVersion: telemetry.openstack.org/v1beta1
...
+      aodh:
+        secret: osp-secret
+        passwordSelectors:
+        databaseUser: aodh
+        databaseInstance: openstack
+        memcachedInstance: memcached
+      heatInstance: heat
+  ceilometer:
+    enabled: true
+    template:
```
jlarriba, line 20: if we use inline declaration in the telemetry_types, this would be:

```
spec:
  ceilometer:
     enabled: true
     secret: osp-secret
...
```

for better readability.

config/samples/telemetry_v1beta1_telemetry.yaml
```diff
@@ -2,8 +2,21 @@ apiVersion: telemetry.openstack.org/v1beta1
...
+      secret: osp-secret
```
jlarriba, line 21: If we deploy the secret in telemetry, we do not declare it in ceilometer.

tests/kuttl/suites/default/tests/01-deploy.yaml
```diff
@@ -4,18 +4,22 @@ metadata:
...
+        listenerImage: "quay.io/mmagr/openstack-aodh-listener:current-podified"
+        passwordSelectors:
+        databaseUser: aodh
+        databaseInstance: openstack
+        memcachedInstance: memcached
+      heatInstance: heat
   ceilometer:
-    secret: osp-secret
+    enabled: true
+    template:
```
jlarriba, line 24: Same as above, spare the "template" declaration.

tests/kuttl/suites/default/tests/01-deploy.yaml
```diff
@@ -4,18 +4,22 @@ metadata:
...
-      evaluatorImage: "quay.io/mmagr/openstack-aodh-evaluator:current-podified"
-      notifierImage: "quay.io/mmagr/openstack-aodh-notifier:current-podified"
-      listenerImage: "quay.io/mmagr/openstack-aodh-listener:current-podified"
-      passwordSelectors:
-      databaseUser: aodh
-      databaseInstance: openstack
-      memcachedInstance: memcached
-    heatInstance: heat
+    enabled: true
+    template:
```
jlarriba, line 8: Same as above, spare the template declaration

config/samples/telemetry_v1beta1_telemetry.yaml
```diff
@@ -2,8 +2,21 @@ apiVersion: telemetry.openstack.org/v1beta1
 kind: Telemetry
 metadata:
   name: telemetry
-  namespace: openstack
 spec:
   autoscaling:
-    prometheus:
-      deployPrometheus: false
+    enabled: false
+    template:
```
jlarriba, line 8: Same as explained for ceilometer.

jlarriba (changes requested):

api/v1beta1/telemetry_types.go
vyzigold: I see your point and I'm not against it. The reason why I did it like this is, that I thought, that users would already be familiar with the openstack control plane CR and my goal was to make the telemetry resource similar to that. In the openstack control plane you'd have:
```
service_name:
  enabled: true or false
  template:
    <the service CR>
service_name2:
  ...
```
Which is what I have here right now. I wonder why cinder decided to go a different way and which of the operators we should follow.

jlarriba: All service operators follow the same pattern. Cinder was just an example. The template explicit field is only used in openstack-operator.

vyzigold: Alright, I'll change it then.

PR #241, opened 2023-11-21
Added Logging configuration to telemetry-operator

jlarriba:


controllers/logging_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
+
+	// ConfigMap
+	configMapVars := make(map[string]env.Setter)
+
+	//
+	// create Configmap required for logging-compute input
+	// - %-config configmap holding minimal logging config required to get the service up, user can add additional files to be added to the service
+	//
+	err = r.generateComputeServiceConfig(ctx, helper, instance, &configMapVars)
```
vyzigold, line 212:
```suggestion
	// Secret
	secretVars := make(map[string]env.Setter)

	//
	// create Secret required for logging-compute input
	// - %-config secret holding minimal logging config required to get the service up, user can add additional files to be added to the service
	//
	err = r.generateComputeServiceConfig(ctx, helper, instance, &secretVars)
```

controllers/logging_controller.go
```diff
@@ -0,0 +1,332 @@
...
+		Log.Info(fmt.Sprintf("Input maps hash %s - %s", common.InputHashName, hash))
+	}
+	return hash, changed, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *LoggingReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&telemetryv1.Logging{}).
+		Owns(&corev1.Secret{}).
```
vyzigold, line 330: In the `reconcileNormal` you are creating a deployment, service and also some rbac rules. Don't you need to own any of these?

vyzigold: I see, that the service is created in a different namespace and you can't own it because of that, but what about the rest?

controllers/logging_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	//
+	// initialize status
+	//
+	if instance.Status.Conditions == nil {
+		instance.Status.Conditions = condition.Conditions{}
+		// initialize conditions used later as Status=Unknown
+		cl := condition.CreateList(
+			condition.UnknownCondition(condition.InputReadyCondition, condition.InitReason, condition.InputReadyInitMessage),
+			condition.UnknownCondition(condition.ServiceConfigReadyCondition, condition.InitReason, condition.ServiceConfigReadyInitMessage),
+			condition.UnknownCondition(condition.DeploymentReadyCondition, condition.InitReason, condition.DeploymentReadyInitMessage),
```
vyzigold, line 139: Are you setting this condition to true (or false) anywhere?

controllers/logging_controller.go
```diff
@@ -0,0 +1,332 @@
...
+			// service account, role, rolebinding conditions
+			condition.UnknownCondition(condition.ServiceAccountReadyCondition, condition.InitReason, condition.ServiceAccountReadyInitMessage),
+			condition.UnknownCondition(condition.RoleReadyCondition, condition.InitReason, condition.RoleReadyInitMessage),
+			condition.UnknownCondition(condition.RoleBindingReadyCondition, condition.InitReason, condition.RoleBindingReadyInitMessage),
+			// right now we have no dedicated KeystoneServiceReadyInitMessage
+			condition.UnknownCondition(condition.KeystoneServiceReadyCondition, condition.InitReason, ""),
```
vyzigold, line 145: Do we need this condition? It doesn't look like you are interacting with keystone at all for logging.

vyzigold (commented):
Wouldn't it be a good idea to add this to the kuttl test suite?

controllers/logging_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *LoggingReconciler) generateComputeServiceConfig(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *telemetryv1.Logging,
+	envVars *map[string]env.Setter,
+) error {
+
```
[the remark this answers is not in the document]
jlarriba: yeah, I was keeping that for emotional reasons: configuration should not go in Secrets. However, it seems that we are not going to go back to use configmaps and that even OpenShift operators are using secrets to store configuration, this is a very good change.

controllers/logging_controller.go
jlarriba: Absolutely true sir. I have also changed the name of the Deployment creation func to DataPlaneDeployment to avoid confusion with the Deployment Kubernetes object.

jlarriba:
> Wouldn't it be a good idea to add this to the kuttl test suite?

Yes indeed, working on it.

jlarriba:
recheck

controllers/logging_controller.go
vyzigold: So what about the service and rbac stuff here? We don't need to own it?

controllers/logging_controller.go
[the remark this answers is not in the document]
jlarriba: Something was lost in the vast seas of rebasing, re-added again.

controllers/logging_controller.go
vyzigold: thanks

jlarriba:
recheck

vyzigold (approved):

PR #242, opened 2023-11-24
Remove secret and related logic from telemetry

vyzigold:


vyzigold:
While this might be a good idea, it's hard to implement proper check for required values if the condition is that "at least one of the secrets is specified". With this PR we can just require the secrets in spec.autoscaling.aodh.secret and spec.ceilometer.secret

jlarriba:
recheck

PR #248, opened 2023-12-14
Metric storage

vyzigold:
This pull request adds a new resource type: metricStorage.

This uses the OBO/COO MonitoringStack as well as ServiceMonitor and ScrapeConfig to deploy everything needed to scrape and store metrics. The user can use the "redhatMetricStorage" field to describe the metric storage with a few options we decided to expose or they can specify their own monitoringStack using the "customMonitoringStack".

Inclusion of the metricStorage into the telemetry resource will follow in a separate PR soon after this is merged.

There are 3 slight issues this is hitting because of the pin of controller-runtime to v0.14, which forces us to also pin OBO to an older version.

- The ScrapeConfig isn't available yet in this pinned OBO version, so we are forced to use the "Unstructured" type to deploy it. This works OK, we are able to deploy the right ScrapeConfig. The only downside other than maybe less readable code is, that the detection of changes when calling CreateOrPatch doesn't seem to work as good. While the ServiceMonitor gets updated once or twice, the ScrapeConfig gets updated like 10 times for the same set of events in the cluster.
- We can't set the default scrapeInterval for MonitoringStack, because it isn't exposed in the CRD at this pinned OBO version. This isn't a big deal though, because we can still set it in the ServiceMonitor and ScrapeConfig and get the same result.
- Because of the above, the user isn't able to set the scrapeInterval when using the "customMonitoringStack", because this option isn't part of our CRD yet. So they're stuck with the default scrapeInterval for now.


pkg/metricstorage/scrape_config.go
```diff
@@ -0,0 +1,67 @@
...
+		"spec": map[string]interface{}{
+			"scrapeInterval": scrapeInterval,
+			"staticConfigs": []interface{}{
+				map[string]interface{}{
+					"targets": targets,
+				},
+			},
+		},
+	})
+	// NOTE: SetUnstructuredContent will overvrite any data, including GVK, name, ...
```
jlarriba, line 56: nit: "overwrite"

vyzigold:
recheck

jlarriba (approved):

PR #256, opened 2023-12-18
Add metricstorage to telemetry

vyzigold:
This PR adds metricstorage to the telemetry api.

I needed to adjust the metricstorage api and change the redhatMetricStorage and the customMonitoringStack to pointers. Without that I wasn't able to make the telemetry controller leave the fields empty (it would always set both of them no matter what was in the telemetry CR applyed by the user). This way I was able to set the empty fields to nil. I adjusted a few checks in the metricstorage accordingly.


vyzigold:
The metricstorage controller is waiting for nodeset CRD to appear before deploying ScrapeConfig. This is expected, but because there is no dataplane-operator installed, the CRD never appears. I tried to add a small "nodeset" CRD, which should work for the purposes of kuttl tests in https://github.com/openstack-k8s-operators/telemetry-operator/pull/258 . If it works there, I'll add it here as well.

vyzigold:
recheck

api/v1beta1/telemetry_types.go
```diff
@@ -95,6 +113,9 @@ type TelemetryStatus struct {
 
 	// ReadyCount of Autoscaling instance
 	AutoscalingReadyCount int32 `json:"autoscalingReadyCount,omitempty"`
+
+	// ReadyCount of MetricStorage instance
+	MetricStorageReadyCount int32 `json:"metricstorageReadyCount,omitempty"`
```
jlarriba, line 118: do we really want to support the use case where we have multiple MetricStorages controlled by this operator? Seems a little bit strange.

config/samples/telemetry_v1beta1_telemetry.yaml
```diff
@@ -3,6 +3,18 @@ kind: Telemetry
 metadata:
   name: telemetry
 spec:
+  metricStorage:
+    enabled: true
```
jlarriba, line 7: lets leave this disabled by default in our examples.

api/v1beta1/metricstorage_types.go
```diff
@@ -89,13 +89,15 @@ type MetricStorageSpec struct {
 	// RedHatMetricStorage allows to define a metric storage with
 	// options supported by Red Hat
 	// +kubebuilder:validation:Optional
-	RedHatMetricStorage RedHatMetricStorage `json:"redhatMetricStorage,omitempty"`
+	// +nullable
+	RedHatMetricStorage *RedHatMetricStorage `json:"redhatMetricStorage,omitempty"`
```
jlarriba, line 93: Now I  am having second thoughts about calling this RedHatMetricStorage, I dont think I like this. Why can't it be just "MetricStorage"?

controllers/telemetry_controller.go
```diff
@@ -341,11 +354,78 @@ func reconcileAutoscaling(ctx context.Context, instance *telemetryv1.Telemetry,
...
+// reconcileMetricStorage ...
+func reconcileMetricStorage(ctx context.Context, instance *telemetryv1.Telemetry, helper *helper.Helper) (ctrl.Result, error) {
+	const (
+		metricStorageNamespaceLabel = "MetricStorage.Namespace"
+		metricStorageNameLabel      = "MetricStorage.Name"
+		metricStorageName           = metricstorage.DefaultServiceName
+	)
+	metricStorageInstance := &telemetryv1.MetricStorage{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      metricstorage.DefaultServiceName,
```
jlarriba, line 366: Previously we try to keep track of the number of MetricStorages we are controlling, but all of them would be called the same. That seems conflicting.

api/v1beta1/telemetry_types.go
vyzigold: You are right. The telemetry CRD isn't made to support multiple MetricStorages, although I was asked to make the metricstorage controller capable of handling multiple metricstorages just in case we want to enable that in the future.

But this line is mostly about following the rest of the resources. Do we support multiple autoscalings or multiple ceilometers? Not sure about ceilometers, but I'm pretty sure we can't have multiple autoscalings in one namespace with the current code.

config/samples/telemetry_v1beta1_telemetry.yaml
vyzigold: yes, agreed

tests/kuttl/suites/default/tests/00-deps.yaml
```diff
@@ -3,4 +3,7 @@ kind: TestStep
 commands:
   - script: |
       oc apply -f ../deps/rhobs.yaml
+      if ! [[ `oc get crd openstackdataplanenodesets.dataplane.openstack.org` ]]; then
```
jlarriba, line 6: this should come from a rebase with master

api/v1beta1/metricstorage_types.go
vyzigold: I don't mind renaming it. But my thought was, that it looked strange and confusing to have "MetricStorage" twice inside each other when deploying telemetry. The CR would look something like this:

```
spec:
  metricStorage:
    enabled: true
    metricStorage:
      type: prometheus
  autoscaling:
    ...
```

api/v1beta1/telemetry_types.go
jlarriba: Yes, but the controller will never be able to handle multiple metricstorages because all of them would have the same name in the same namespace. I would recommend to fix this for metricstorage now and fix that for autoscaling and ceilometer in a future PR.

api/v1beta1/metricstorage_types.go
jlarriba: we can always call the first one "telemetryStorage"

vyzigold: well, currently the top fields in telemetry spec are named after the resource they belong to. So "autoscaling" for the autoscaling resource, ceilometer for ceilometer and metricStorage for metricStorage. I quite like this, I think it tells the user what they're configuring. Maybe we could rename the second one? (rename RedHatMetricStorage to telemetryStorage)

controllers/telemetry_controller.go
vyzigold: It's the same for ceilometer and autoscaling in the telemetry controller

api/v1beta1/metricstorage_types.go
jlarriba: what if we call the top level simply "storage"? The difference between this one and ceilometer and autoscaling is that under this one we can configure two "different" resources, which does not happen on the others. So I would say that this is a special one.

api/v1beta1/telemetry_types.go
vyzigold: So the suggestion is to remove all of the *ReadyCount from the status?

vyzigold: I guess the *Ready conditions should be enough, so why not

jlarriba: well, i other operators this field is used to inform about the number of replicas of the service that are running, but we are fixing the number of replicas of all of our services to 1. What would be better, use the same as the rest of the operators or replace this count with a bool that says if the 1 replica is ready? I dont know, maybe for the moment we can leave it like this and decide later.

jlarriba: we crossed. Right, the Ready conditions should be enough.

api/v1beta1/metricstorage_types.go
vyzigold: Hmm, I'm not sure if we understand each other on where the 2 different resources are configured. Not taking into account any renaming right now, the 2 cases of deploying the different resources from the point of view of telemetry would look like this:
```
spec:
  metricStorage:
    enabled: true
    redhatMetricStorage:
      type: prometheus
  <autoscaling and ceilometer>
```
```
spec:
  metricStorage:
    enabled: true
    customMonitoringStack:
      logLevel: debug
  <autoscaling and ceilometer>
```
In both cases the telemetry controller will take the contents of metricStorage and create a new resource in the cluster of kind "MetricStorage" and name "metric-storage". The top level in the telemetry CR doesn't distinguish between the different resources.

api/v1beta1/telemetry_types.go
vyzigold: ok, I'll remove the counts then

api/v1beta1/metricstorage_types.go
jlarriba: I think we understand each other. I dont like the "redhatMetricStorage" there, it should be something different so I was proposing ideas to change either that one to something different or the top level "metricStorage" to something different so redhatMetricStorage could be metricStorage. Maybe "redhatMetricStorage" can be "monitoringStack".

vyzigold: We took the conversation to slack and decided to rename the "redhatMetricStorage" to "monitoringStack". This will be a part of a following PR as it's out of topic here

PR #258, opened 2023-12-20
Add metricstorage kuttl testing

vyzigold:
I didn't previously notice, that the metricstorage kuttl tests actually aren't being executed. This PR adds metricstorage kuttl tests to the kuttl-test.yaml file and fixes the tests to make them succeed.

### DIFF
--- a/tests/kuttl/suites/metricstorage/tests/00-deps.yaml
+++ b/tests/kuttl/suites/metricstorage/tests/00-deps.yaml
@@ -3,4 +3,7 @@ kind: TestStep
 commands:
   - script: |
       oc apply -f ../deps/rhobs.yaml
+      if ! [[ `oc get crd openstackdataplanenodesets.dataplane.openstack.org` ]]; then
+        oc apply -f ../deps/dataplane.openstack.org_openstackdataplanenodesets.yaml
+      fi
       until oc api-resources | grep -q rhobs; do sleep 1; done
--- a/tests/kuttl/suites/metricstorage/tests/01-assert.yaml
+++ b/tests/kuttl/suites/metricstorage/tests/01-assert.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   labels:
-    prometheus: telemetry-kuttl-prometheus
+    prometheus: telemetry-kuttl
   name: prometheus-telemetry-kuttl-0
 status:
   containerStatuses:
@@ -37,13 +37,11 @@ metadata:
     service: metricStorage
   name: telemetry-kuttl
   ownerReferences:
-    kind: MetricStorage
+  - kind: MetricStorage
     name: telemetry-kuttl
 spec:
   endpoints:
-  - bearerTokenSecret:
-      key: ""
-    interval: 30s
+  - interval: 30s
     metricRelabelings:
     - action: labeldrop
       regex: pod
@@ -60,15 +58,16 @@ spec:
     matchLabels:
       service: ceilometer
 ---
-apiVersion: monitoring.rhobs/v1
+apiVersion: monitoring.rhobs/v1alpha1
 kind: ScrapeConfig
 metadata:
   labels:
     service: metricStorage
   name: telemetry-kuttl
   ownerReferences:
-    kind: MetricStorage
+  - kind: MetricStorage
     name: telemetry-kuttl
 spec:
+  scrapeInterval: 30s
   staticConfigs:
-  - targets:
+  - targets: []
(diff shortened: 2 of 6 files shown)


tests/kuttl/suites/metricstorage/tests/01-assert.yaml
```diff
@@ -60,15 +58,16 @@ spec:
     matchLabels:
       service: ceilometer
 ---
-apiVersion: monitoring.rhobs/v1
+apiVersion: monitoring.rhobs/v1alpha1
```
csibbitt, line 61: This looks suspicious to me. This isn't directly related to this PR, and I'm not saying you need to move immediately, but STF is already using the v1 API. You should probably move soon so that things remain stable for you and better aligned between the projects.

I see you are using github.com/rhobs/observability-operator v0.0.20[1], and relying on the v1alpha1 API explicitly[2].

For comparison, here's what I see in my STF deployment (v0.0.26 and v1 API)
```
$ oc get csv | grep observ
observability-operator.v0.0.26                 Cluster Observability Operator           0.1.0                                                          Succeeded

$ oc get apiservices | grep rhobs
v1.monitoring.rhobs                           Local                                                        True        92d
```

[1] https://github.com/openstack-k8s-operators/telemetry-operator/blob/dc4a70dd8059b0331d736c50d304801fb53cb3d5/go.mod#L23
[2] https://github.com/openstack-k8s-operators/telemetry-operator/blob/dc4a70dd8059b0331d736c50d304801fb53cb3d5/main.go#L48

csibbitt (approved):
I think we're waiting on "ci/prow/telemetry-operator-build-deploy-kuttl"[1]  to pass in order to prove that this works? I'm new to this codebase and haven't looked deeper, just assuming based on the names.

In general this LGTM

[1] https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_telemetry-operator/258/pull-ci-openstack-k8s-operators-telemetry-operator-main-telemetry-operator-build-deploy-kuttl/1737516578842349568

tests/kuttl/suites/metricstorage/tests/01-assert.yaml
vyzigold: Right now, v1alpha1 is the only version, which has ScrapeConfig. Look at [1], which is where it lives for OBO. There is no ScrapeConfig in the other versions.

Regarding the OBO versions, we use v0.0.20 in code to get the golang structs. We are currently stuck with this version, because of other dependencies like controller-runtime or k8s.io/* . OBO in newer versions uses newer versions of those dependencies. Unfortunately it's not trivial to update these dependencies for us as it requires quite a bit of code changes and all the openstack operators must do this at the same time. But fortunately there are discussions ongoing to do the update rather sooner than later.

v0.0.20 is just the version we currently use to get the golang structs at the moment. For testing and development we've been using [2] to deploy OBO into the cluster. Which at the moment installs v0.0.29 for me.

[1] https://github.com/rhobs/obo-prometheus-operator/blob/rhobs-rel-0.69.1-rhobs1/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
[2] https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/tests/kuttl/suites/metricstorage/deps/rhobs.yaml

vyzigold:
The metricstorage controller is waiting for nodeset CRD to appear before deploying ScrapeConfig. This is expected, but because there is no dataplane-operator installed, the CRD never appears. I tried to add a small "nodeset" CRD, which should work for the purposes of kuttl tests.

tests/kuttl/suites/metricstorage/tests/00-deps.yaml
```diff
@@ -3,4 +3,7 @@ kind: TestStep
 commands:
   - script: |
       oc apply -f ../deps/rhobs.yaml
+      if ! [[ `oc get crd openstackdataplanenodesets.dataplane.openstack.org` ]]; then
```
jlarriba, line 6: Why are you checking the if the nodesets crd is on the cluster before applying? None of the other kuttl tests in this operator is applying this, so it should never be there once the kuttl tests run on the CI, right?

vyzigold: Yes, it would work in CI even without the check. But what if you want to run the tests locally in your environment? In the future, if the nodeset CRD changes, this would overwrite that CRD with the one we have saved in the tests. This might break your environment. That's why I decided to check.

PR #260, opened 2023-12-21
Move ceilometer to use StatefulSet

vyzigold:
This moves ceilometer to use StatefulSet instead of deployment. This aligns us to the rest of the operators and also allows us to check for ceilometer pod in kuttl tests (the tests to check for the pod is included in this PR)

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -35,6 +35,18 @@ rules:
   - update
   - watch
 - apiGroups:
+  - apps
+  resources:
+  - statefulsets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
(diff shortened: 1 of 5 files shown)


vyzigold:
This needs an adjustment to the default kuttl test suite. There is also this PR: https://github.com/openstack-k8s-operators/telemetry-operator/pull/256 Which also makes changes tot he default kuttl suite. I'll wait until that one merges and then continue working on this.

vyzigold:
The PR is rebased on top of main and the required modification to default kuttl test suit is also there. If the CI passes, it's ready for reviews.

jlarriba (approved):
/lgtm

PR #263, opened 2023-12-22
Move autoscaling to use StatefulSet

vyzigold:
Similarly to https://github.com/openstack-k8s-operators/telemetry-operator/pull/260 this moves autoscaling to use statefulset for aodh instead of deployment. This aligns us to the rest of the operators and also allows us to check for aodh pod in kuttl tests

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -25,18 +25,6 @@ rules:
 - apiGroups:
   - apps
   resources:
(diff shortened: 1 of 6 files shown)


vyzigold:
I just had to rebase on top of main and execute `make generate` `make manifests` to regenerate the rbac. Nothing else changed, so I readded the lgtm label.

PR #269, opened 2024-01-04
update CSV yaml with MetricStorage

yazug:


vyzigold:
The CI issue was caused by a patch, which is reverted here: https://github.com/openshift/release/pull/47407 . There are apparently other issues, so the CI isn't expected to work for now.

PR #271, opened 2024-01-10
Add default prometheus port to consts

vyzigold:


vyzigold:
I haven't run the operator with these changes yet. I'll cancel the hold when I do

PR #280, opened 2024-01-18
Update ceilometer compute agent template

vyzigold:


templates/ceilometercompute/config/ceilometer_agent_compute.json
```diff
@@ -1,6 +1,6 @@
 {
     "image": "{{ .ceilometer_compute_image }}",
-    "user": "root",
+    "user": "ceilometer",
```
jlarriba, line 3: Does the user Ceilometer have enough permissions to connect to libvirt socket?

vyzigold: We can't connect to the libvirt socket and it doesn't matter if we use root or ceilometer user at the moment. Check https://issues.redhat.com/browse/OSPRH-3373 for more context. I haven't seen any additional issues when switching the user to ceilometer. When using the WA from the issue or switching SELinux to permissive, I can confirm that ceilometer has enough permissions.

jlarriba:
/retest

There is a timeout connecting quay.io, it seems to be a temporary error, as it now works fine.

csibbitt (approved):

paramite (approved):

PR #286, opened 2024-01-25
s/remove/delete/ in README.md

csibbitt:
```
$ oc remove
error: unknown command "remove" for "oc"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ oc patch openstackcontrolplane openstack-galera-network-isolation --type='json' 
 7.- Remove telemetry-operator from the deployments
 ```
 oc project openstack-operators
-oc remove csv telemetry-operator.v0.0.1
+oc delete csv telemetry-operator.v0.0.1
 ```
 
 8.- Deploy custom telemetry-operator version


README.md
```diff
@@ -46,15 +46,15 @@ make edpm_nova_discover_hosts
 
 Now, we proceed to run our own telemetry-operator instance:
 
-6.- Remove Ceilometer deployment
+6.- Remove Telemetry deployment
 ```
-oc patch openstackcontrolplane openstack-galera-network-isolation --type='json' -p='[{"op": "replace", "path": "/spec/ceilometer/enabled", "value":false}]'
+oc patch openstackcontrolplane openstack-galera-network-isolation --type='json' -p='[{"op": "replace", "path": "/spec/telemetry/enabled", "value":false}]'
```
csibbitt, line 51: I'm less sure about this one. Was this from an earlier time when you were dealing only with ceilometer, or have I misunderstood? If this is enabled, the openstack operator controls a telemetry object, but we want to deploy our own (including the embedded ceilometer definition) in Step 9

README.md
```diff
@@ -65,7 +65,7 @@ oc delete -f config/crd/bases/
 oc apply -f config/crd/bases/
 
 make manifests generate
-OPERATOR_TEMPLATES=$PWD/templates make run
+OPERATOR_TEMPLATES=$PWD/templates make run-with-webhook
```
csibbitt, line 68: If I use the old version (`make run`), I cannot apply telemetry manifests, getting this error:

```
$ oc apply -f config/samples/telemetry_v1beta1_telemetry.yaml
The Telemetry "telemetry" is invalid:                                                       
* spec.autoscaling.aodh.apiImage: Required value       
* spec.autoscaling.aodh.evaluatorImage: Required value 
* spec.autoscaling.aodh.listenerImage: Required value                                                                                                                                    
* spec.autoscaling.aodh.notifierImage: Required value  
* spec.ceilometer.centralImage: Required value                                                                                                                                           
* spec.ceilometer.computeImage: Required value 
* spec.ceilometer.ipmiImage: Required value
* spec.ceilometer.nodeExporterImage: Required value
* spec.ceilometer.notificationImage: Required value
* spec.ceilometer.sgCoreImage: Required value
```

The necessary webhook configurations are destroyed when we delete the CSV in step 7, so there is nothing to set the defaults on new telemetry manifests.

README.md
vyzigold: I'm a little unsure about the usefulness of this line. I don't think it ever worked for me as intended. If you disable telemetry in oscp, the openstack-operator will just delete any telemetry object you create, so you can't test anything. @jlarriba is this actually useful at all?

README.md
vyzigold: I don't mind changing the line to run-with-webhook. You can still use `make run` if you make a copy of that CR and fill in the required values. They shouldn't be put into the samples though.

README.md
jlarriba: It is useful as it removes all the objects created by openstack-operator. When a development version of T-O is then deployed, this enabled can be changed to true to get things deployed by the new version of T-O. In a certain moment of the dev cycle, the objects created by the "old" version of T-O need to be removed from the cluster, and this seems to be the most straightforward way.

README.md
jlarriba: In the tests I did in the distant past (this could have changed ofc), the webhooks did not work with the T-O running in "dev mode" because it does not expose the correct endpoints on the cluster for the webhook to use. If that's still the case, I am against running with webhooks, of course, because it will not work. Jaromir and me share versions of the sample manifests with the images filled in (that cannot go in the samples dir) because of this. We can still share those with the rest of the team to end up with the issue you are highlighting here.

@csibbitt has the issue with webhooks has been resolved for local env? It seemed pretty structural months ago.

jlarriba: @csibbitt Just say your Jira comment. If it works correctly, this change is good for me.

README.md
csibbitt: Yes, I was using it as Juan described, to remove the openstack deployed artifacts and then either:
1. Bring them back by re-enabling this  -or-
2. Scale down the openstack operator to prevent any more interference

README.md
csibbitt: Ah yes, the manifests in our wiki about testing autoscaling have the images filled in and don't rely on the mutating webhook, which is why I was able to follow those instructions top-to-bottom without trouble. These README instructions just plain didn't work, though. Maybe I could add a note to the instructions here that clarify.

README.md
csibbitt: This is the command for (2) above - `oc scale deploy/openstack-operator-controller-manager --replicas=0 -n openstack-operators`

If following the current instructions, Step 9 is going to fail because the openstack operator will remove the telemetry object you apply, as Jaromir pointed out. So I think I really should add this step to disable the openstack operator here.

jlarriba: Yes if you apply a new one, but what I usually end up doing to prevent removing the whole openstack-operator is to enable this one again once my T-O instance is running instead of applying a new one that is, most of the time, identical as the one openstack-operator would be applying if re-enabled.

However, I understand the README creates confusion as it will not work as it is. Dont know if recommending deleting the openstack-operator is something we want to do generally as most of the time enabling/disabling this would do.

README.md
jlarriba: Please do.

README.md
csibbitt: Okay, in that case I suggest we keep this how it is and replace Step 9 with two alternatives "If you'd like to continue running openstack-operator and use its telemetry object, `oc patch...telemetry/enabled...true`. If you'd like to deploy a custom telemetry object `oc scale...replicas=0; oc apply -f...`"

Does that make sense?

jlarriba: Perfect sense to me, thank you.

jlarriba:
/retest

Changes in the README cannot affect the kuttl tests

PR #287, opened 2024-01-26
[go.mod]Bump from v0.3.0 to pseudoversion

gibizer:
Our intention is to track service operator and lib-common dependencies via pseudoversions. However renovate does not automatically bump from a tagged version (e.g. v0.3.0) to a newer but not tagged pseudoversion. So this patch does the manual bump. After this renovate will bump the newer pseduoversions.


vyzigold:
gotidy...................................................................Failed
- hook id: gotidy
- files were modified by this hook

jlarriba:
There it seems to be an issue with the CI, both the github-check and the kuttl deploy are failing due to the same issue:

++ /bin/bash hack/pin-bundle-images.sh
jq: error (at <stdin>:1): Cannot iterate over null (null)

abays:
Hmm, I wonder if the problem with the build is that @gibizer 's Telemetry Operator repo is called `ceilometer-operator` instead of `telemetry-operator`?  Maybe that confuses the build script, which assumes the remote forks have the same repo names `openstack-k8s-operators`?

vyzigold:
Seems like I've hit the same error in my recent openstack-operator PR as well, so gibizer's repo name may not be the cause: https://github.com/openstack-k8s-operators/openstack-operator/pull/643

abays:
> Seems like I've hit the same error in my recent openstack-operator PR as well, so gibizer's repo name may not be the cause: [openstack-k8s-operators/openstack-operator#643](https://github.com/openstack-k8s-operators/openstack-operator/pull/643)

Ah, okay.  I think you're probably right then and the repo name variance isn't to blame.

jlarriba:
This PR has been submitted to all other operators and it is failing on all of them. Apparently, it seems to be related with a change in mariadb-operator. It is being debugged in https://github.com/openstack-k8s-operators/nova-operator/pull/660

jlarriba:
recheck

Let's try again

jlarriba:
The same as all the other times:

```
error=\"no csv in bundle\"\nError: no csv in bundle\
...
~/src/github.com/openstack-k8s-operators/openstack-operator ~/ci-framework-data/artifacts", "jq: error (at <stdin>:1): Cannot iterate over null (null)", "# FIXME: hardcoded bundle below should use go.mod pinned version for manila bundle"

```
It seems to be a very specific error regarding manila bundle not having the csv

gibizer:
I did a rebase that picked up some go.sum changes maybe this helps.

gibizer:
There is some telemetry specific build error I don't really get.

PR #294, opened 2024-02-06
Controller runtime update

bshephar:
This changes updates the controller-runtime and fixes the initialisation of objects to comply with the new controller-runtime standards. Namely:

We fix the function signatures of the webhook to include admissions.Warnings:
https://github.com/kubernetes-sigs/controller-runtime/commit/8770b4d3b5425b156f48c971ee140ea25f9c1bfa

We fix the initialisation of the manager to declare the webhook following the new schema:
https://github.com/kubernetes-sigs/controller-runtime/commit/91cdd8c9539c4860c5bacad36ebf75e4271f726d

We fix the Watches function to comply with the new signature definition:
https://github.com/kubernetes-sigs/controller-runtime/commit/dc62ed5dadc6ea2d3e22546368b939a2db1edfb2

Add context to EventHandlers:
https://github.com/kubernetes-sigs/controller-runtime/commit/2464a9d7af4911c6bf6e0baaaf8dc0ab08c7452c


controllers/metricstorage_controller.go
```diff
@@ -182,6 +183,14 @@ func (r *MetricStorageReconciler) reconcileNormal(
 	Log := r.GetLogger(ctx)
 	Log.Info(fmt.Sprintf("Reconciling Service '%s'", instance.Name))
 
+	var mgr manager.Manager
```
bshephar, line 186: We need to use the manager we initialise in main.go rather than a new one. So, maybe we add mgr to the `MetricStorageReconciler` struct so that it can be used in these two functions.

I just did this while I was figuring out what worked.

controllers/metricstorage_controller.go
```diff
@@ -399,13 +398,12 @@ func (r *MetricStorageReconciler) ensureWatches(
...
 		Name: name,
 	}, u)
-
 	if err != nil {
 		return err
 	}
 
 	Log.Info(fmt.Sprintf("Starting to watch %s", name))
-	err = r.Controller.Watch(&source.Kind{Type: kind},
+	err = r.Controller.Watch(&source.Kind{Type: kind}, // TODO(bshephar) This is no longer correct with the CR update. Needs to be fixed.
```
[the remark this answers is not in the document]
bshephar: Fixed

controllers/metricstorage_controller.go
bshephar: Using the Reconciler struct. It's either that, or we push them to the heap in their own variables during `SetupWithManager`

api/go.mod
```diff
@@ -1,75 +1,77 @@
...
-go 1.19
+go 1.21
+
+toolchain go1.21.1
 
 require (
 	github.com/onsi/ginkgo/v2 v2.14.0
 	github.com/onsi/gomega v1.30.0
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.3.1-0.20240124141114-55d029e4658b
 	github.com/rhobs/observability-operator v0.0.20
```
vyzigold, line 11: This is a bit surprising to me, OBO wasn't bumped too. I remember I needed to use older version, because higher versions required higher versions of controller-runtime. If it runs with this version, then it's OK to leave it like this for now. There will be a bunch of more code changes once we bump this.

vyzigold (approved):
/lgtm

I tried running it and I can confirm it runs. There will need to be a bunch of changes in the MetricStorage once we bump OBO, which should be soon now. We can check the changes around the watch functions together with that.

api/go.mod
csibbitt: I came looking for this

csibbitt (commented):
LGTM but I don't have the full context to approve

bshephar:
We need to fix this before we can merge:
https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/controllers/metricstorage_controller.go#L523-L532

Services list no longer exists on the `OpenStackDataPlaneService` and was removed by:
https://github.com/openstack-k8s-operators/dataplane-operator/commit/23868fbb15d981acdaaa81d05acb011a0c395c71

We can get the reserved IP address of the node by looking at the `OpenStackDataPlaneNodeSet`:
```
[m3@localhost ~]$ oc get osdpns -o yaml | yq '.items[].status.AllIPs'
{
  "edpm-compute-0": {
    "ctlplane": "192.168.122.100",
    "internalapi": "172.17.0.100",
    "storage": "172.18.0.100",
    "tenant": "172.19.0.100"
  }
}
```

We should probably refactor this section and decide how to best achieve the desired outcome for this use case.

PR #301, opened 2024-02-15
[zuul] Add autoscaling job

elfiesmelfie:
This job will deploy openstack with autoscaling components enabled

* Use a pre_deploy hook to create the COO subscription

* Pass kustomization for autoscaling to the tempest job

PR #302, opened 2024-02-16
DNM testing addmetricsstorage

elfiesmelfie:
This is a temporary PR to test re-anabling metrics storage with alertmanage disabled. It is being tested separate so I can test that the alertmanager was the issue and move with some other update to the main PR #301 

Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/301


elfiesmelfie:
The jobs failed, but ran far enough for me to be confident that alert manager is no longer enabled.
This change has been cherry-picked onto the branch for PR301

PR #284, opened 2024-01-22
[zuul] Add telemetry-operator-multinode-autoscaling-tempest job

elfiesmelfie:
This is a first pass, creating a tempest job based on neutron-operator. This change runs the neutron tempest tests to establish that the environment is configured correctly for tempest tests.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1110
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1116
Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/132
Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/301
Depends-On: https://review.opendev.org/c/openstack/releases/+/907406
Depends-On: https://review.rdoproject.org/r/c/rdoinfo/+/51721

PR #310, opened 2024-02-19
[tlse] tls support for telemetry aodh

Deydra71:
Public/Internal service cert secrets and the CA bundle secret can be passed to configure httpd virtual hosts for tls termination. The certs get direct mounted to the appropriate place in etc/pki/tls/certs/%s.crt|key and a CA bundle to /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem . Job deployments for bootstrap/cron get the CA bundle added if configured.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/428

Jira: https://issues.redhat.com/browse/OSPRH-4197


api/v1beta1/autoscaling_types.go
```diff
@@ -135,6 +135,11 @@ type AutoscalingSpec struct {
 	// Heat instance name.
 	// +kubebuilder:default=heat
 	HeatInstance string `json:"heatInstance"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// TLS - Parameters related to the TLS
+	TLS tls.API `json:"tls,omitempty"`
```
vyzigold, line 142: I think it would make a bit more sense inside the Aodh struct above, since this is configuring TLS for aodh.

vyzigold (commented):
I know this is still WIP, but I took a brief look. So far it looks good except for that one comment I left.

I have a question regarding the CA certs mounted to the containers. We have a bug either in the operator or in aodh. Because of that aodh seems to be ignoring its configuration and it's always using public endpoint for keystone. So the question is if the CA certs are the same for internal and public endpoints? Will aodh be able to connect to public keystone endpoint after we're done with the TLS efforts?

api/v1beta1/autoscaling_types.go
Deydra71: We are using two CAs - one for public and one for internal endpoints, and the `combined-ca-bundle.pem` file  incorporates the CAs for both public and internal endpoints. There's also the possibility to provide service specific CA through `ca.crt`.

So, the aodh should be able to connect to the public keystone.

For more info --> https://github.com/openstack-k8s-operators/docs/blob/main/tls.md

Deydra71: cc: @stuggi

Deydra71:
Currently the volumes and volume mounts in both ceilometer and aodh are not correctly created. Working on it.

vyzigold:
Looks like this is hitting the same issue as Emma's and mine PR, where the infra-operator resources are trying to pull incorrect images: https://logserver.rdoproject.org/10/310/15348e87af91e03038ba3320535d3553e9ace717/github-check/podified-multinode-edpm-deployment-crc/89a8845/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/memcached-0/memcached-0-describe

vyzigold:
We had the same issue here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/284 . It should be fixed, so next run should get the CI further.

elfiesmelfie:
> We had the same issue here: #284 . It should be fixed, so next run should get the CI further.

The issue is resolved in https://github.com/openstack-k8s-operators/openstack-operator/pull/643

config/samples/telemetry_v1beta1_tls.yaml
```diff
@@ -0,0 +1,35 @@
...
+    passwordSelectors:
+    databaseUser: aodh
+    databaseInstance: openstack
+    memcachedInstance: memcached
+    tls:
+      api:
+        internal:
+          secretName: cert-telemetry-internal-svc
+        public:
+          secretName: cert-telemetry-public-svc
```
vyzigold, line 34: I'm not sure about the secretNames. The services for aodh are called "aodh-internal" and "aodh-public". And there will be other services telemetry, which will need certificate secrets (I might get a draft PR ready today with that). It would make sense to me to substitute "telemetry" for "aodh".

tests/kuttl/suites/tls/tests/01-assert.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cert-telemetry-internal-svc
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cert-telemetry-public-svc
```
vyzigold, line 9: Same thing with the secret names here. I think it should be cert-aodh-*-svc and in the future we can add other secrets.

controllers/autoscaling_controller.go
```diff
@@ -737,5 +777,43 @@ func (r *AutoscalingReconciler) SetupWithManager(ctx context.Context, mgr ctrl.M
...
 		Complete(r)
 }
+
+func (r *AutoscalingReconciler) findObjectsForSrc(ctx context.Context, src client.Object) []reconcile.Request {
+	requests := []reconcile.Request{}
+
+	l := log.FromContext(context.Background()).WithName("Controllers").WithName("Autoscaling")
+
+	for _, field := range allWatchFields {
+		crList := &telemetryv1.CeilometerList{}
```
vyzigold, line 794:
```suggestion
		crList := &telemetryv1.AutoscalingList{}
```

Deydra71:
@vyzigold
Can you please check the current files for tls kuttl test, and guide me what steps need to be done in order to run tls kuttl? Right now there's only the certificates creation, nonetheless when I run the whole stack of kuttl tests, the tls one is not detected at all.

Thanks!

vyzigold:
I believe you need to edit this file as well: https://github.com/openstack-k8s-operators/telemetry-operator/blob/0d4f1ece1c02a7f30179a09efbc8da2c70ce5563/kuttl-test.yaml#L27 . I hope that's enough to get them executed.

Deydra71:
> I believe you need to edit this file as well:
> 
> https://github.com/openstack-k8s-operators/telemetry-operator/blob/0d4f1ece1c02a7f30179a09efbc8da2c70ce5563/kuttl-test.yaml#L27
> 
> . I hope that's enough to get them executed.

Yep, that was it. Thanks a lot!

vyzigold:
I read through the PR once again and I don't have anything to add. Thank you.
/approve
/lgtm

PR #312, opened 2024-02-21
[zuul] Add a project template for telemetry-tempest jobs

elfiesmelfie:
This will run jobs against changes to telemetry-tempest-plugins The jobs can be triggered with "check rdo experimental" comment

The template will be added to the project in: https://review.rdoproject.org/r/c/config/+/51769

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -29,15 +29,12 @@
     name: rdo-telemetry-tempest-plugin-jobs
     openstack-experimental:
       jobs:
-        - noop:
-            required-projects:
-              - openstack/telemetry-tempest-plugin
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
@jlarriba @vyzigold , this change is to allow testing and then running a job against changes to telemetry-tempest-plugins, so that we can check the compatability with OSP 18 and OSP17 (eventually).
It will trigger on comments to telemetry-tempest-plugin, "rdo check experimantal", or something similar.

This change does nothing right now, so is safe to merge.
It is given power with this change: https://github.com/openstack-k8s-operators/telemetry-operator/pull/312, which will connect it to the telemetry-tempest-plugin changes

jlarriba (approved):

PR #318, opened 2024-02-27
NODE_EXPORTER image here was not used in the compute nodes ansible

jlarriba:


jlarriba:
/unhold

Actually, this is correct and the operator should not be able to define the compute image: that's dataplane-operator job

abays (approved):
/lgtm

PR #319, opened 2024-02-28
Adding Console UI Dashboards automation

csibbitt:
- This depends on the console-dashboards-plugin being installed w/ Helm as described here: https://github.com/openshift/console-dashboards-plugin?tab=readme-ov-file#deployment-on-cluster
- Creates Prometheus recording rules (equivalent to ones used in OCP) to drive the dashboards
- Creates a console UI datasource that points to our prometheus (See: https://github.com/openshift/console-dashboards-plugin/blob/main/docs/add-datasource.md)
- Creates Dashboards (See: https://github.com/openshift/console-dashboards-plugin/blob/main/docs/prometheus-dashboard-example.yaml)


controllers/metricstorage_controller.go
```diff
@@ -265,6 +268,95 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+			Namespace: "console-dashboards",
+		},
+	}
+	op, err = controllerutil.CreateOrPatch(ctx, r.Client, datasourceCM, func() error {
+		datasourceCM.ObjectMeta.Labels = map[string]string{
+			"console.openshift.io/dashboard-datasource": "true",
+		}
+		datasourceCM.Data = map[string]string{
+			// *CS Should template this to prevent problems with the spaces vs. tabs
+			"dashboard-datasource.yaml": `
```
csibbitt, line 319: Even though this looks like a k8s object, it's not, it's a grafana datasource config file. These lines are indented with spaces because `<tab>` doesn't work in the yaml.

controllers/metricstorage_controller.go
```diff
@@ -265,6 +268,95 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+	if op != controllerutil.OperationResultNone {
+		Log.Info(fmt.Sprintf("Prometheus Rules %s successfully changed - operation: %s", prometheusRule.Name, string(op)))
+	}
+	instance.Status.Conditions.MarkTrue(telemetryv1.PrometheusRuleReadyCondition, condition.ReadyMessage)
+
+	// Deploy Configmap for Console UI Datasource
+	// *CS Can I move this to a .Watches callback Fn cross-namespace? (cross-namespace owner references are disallowed)
+	datasourceCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "openstack-" + instance.Name + "-datasource",
```
vyzigold, line 309: I believe we should support cases, where there are multiple control-planes, each in a different namespace. This name as well as the data in the CM should take that into account. Maybe the name should change to
```
instance.Namespace + instance.Name + "-datasource"
```

controllers/metricstorage_controller.go
```diff
@@ -265,6 +268,95 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+			// *CS Should template this to prevent problems with the spaces vs. tabs
+			"dashboard-datasource.yaml": `
+                kind: "Datasource"
+                metadata:
+                    name: "openstack-prometheus-datasource"
+                spec:
+                    plugin:
+                        kind: "PrometheusDatasource"
+                        spec:
+                            direct_url: "http://prometheus-operated.openstack.svc.cluster.local:9090"`,
```
vyzigold, line 327: Same here, we shouldn't rely on the control-plane being deployed into the "openstack" namespace

controllers/metricstorage_controller.go
```diff
@@ -265,6 +268,95 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if op != controllerutil.OperationResultNone {
+		Log.Info(fmt.Sprintf("Prometheus Rules %s successfully changed - operation: %s", prometheusRule.Name, string(op)))
+	}
+	instance.Status.Conditions.MarkTrue(telemetryv1.PrometheusRuleReadyCondition, condition.ReadyMessage)
+
+	// Deploy Configmap for Console UI Datasource
+	// *CS Can I move this to a .Watches callback Fn cross-namespace? (cross-namespace owner references are disallowed)
```
vyzigold, line 306: So, I don't know the answer to this, but I think it would at least be worth a try.

pkg/dashboards/openstack-cloud.go
```diff
@@ -0,0 +1,1061 @@
...
+						"panels": [
+							{
+								"aliasColors": {
+		
+								},
+								"bars": false,
+								"dashLength": 10,
+								"dashes": false,
+								"datasource":  {
+									"name": "openstack-prometheus-datasource",
```
vyzigold, line 70: In an earlier comment I mentioned the datasource name should change with namespace. All these values in this file would need to change too.

controllers/metricstorage_controller.go
vyzigold: I agree that the templating may be a little better. I think the templating process you can see in our other controllers together with configmap.EnsureConfigMaps should create the same ConfigMap.

pkg/dashboards/openstack-cloud.go
vyzigold: Maybe this configmap together with changing the values as mentioned in the first comment could be done with a template too.

vyzigold (commented):
Nice work so far. I went through it and left a few comments.

controllers/metricstorage_controller.go
csibbitt: configmap.EnsureConfigMaps! That's what I was looking for, thanks!

csibbitt: configmap.EnsureConfigMaps ! Thank you, that's what I was looking for yesterday and couldn't put my finger on it.

controllers/metricstorage_controller.go
csibbitt: Yes, thank you. Late in the day I realized I'd hardcoded some things that would make this single-instance only, but I was focused on getting _something_ working. I was hung up on the fact that OBO always creates "prometheus-operated" and we aren't asking it to name it in an instance-specific way, but now with fresh eyes it's obvious that simply templating the namespace portion of the prometheus URL is going to do what we need here.

controllers/metricstorage_controller.go
jlarriba: If we get to have some kind of control over resources created in a different namespace, that would be uttery useful for logging (it creates a Service in the openshift-logging namespace), but until now, I have been stopped by a wall: Kubernetes does not allow for an operator to own resources in a different namespace, so the best I could find was shoot-and-forget.

controllers/metricstorage_controller.go
jlarriba: We tend to store everything on a Secret, so it might be worth to try secret.EnsureSecrets instead of configmaps.

vyzigold: hmm, depends if console UI is able to consume secrets instead of configmaps

pkg/dashboards/openstack-cloud.go
csibbitt: As a side note, in theory I can write this chunk in just one place (instead of 9) as a dashboard template variable, but I've been unable to get it to work so far after significant efforts. I copied what is done in the other dashboards, but since they use the default datasource (which the dashboards appear to revert back to if there is a problem with the definition), I'm not convinced the examples are correct. I tried some variations (putting the whole object in the template value instead of just the name, trying non-obvious places to put the name, etc). Hopefully I can figure this out when I actually fire up Grafana 6 and try to make a dashboard from scratch; in the meantime I'll have to template this value in every panel.

controllers/metricstorage_controller.go
csibbitt: Secrets will not work, the console UI is looking for configmaps.

Unfortunately, I just put in the effort to convert this to use configmap.EnsureConfigMaps and the templating system (though I have some reservations that it might not be well suited), but it doesn't work anyways. Here is the error I get ("cross-namespace owner references are disallowed"):

```
2024-02-28T14:25:04-05:00       ERROR   Reconciler error        {"controller": "metricstorage", "controllerGroup": "telemetry.openstack.org", "controllerKind": "MetricStorage", "MetricStorage": {"name":"metric-storage","namespace":"openstack"}, "namespace": "openstack", "name": "metric-storage", "reconcileID": "8b97084d-697c-47fa-a018-47621a746d45", "error": "error create/updating configmap: cross-namespace owner references are disallowed, owner's namespace openstack, obj's namespace console-dashboards"}
```

I might use some native golang templating to eliminate some of the awkwardness here, but following the usual openstack-k8s-operators pattern is unfortunately not possible.

csibbitt: If it's any consolation, this datasource + dashboard functionality is supposed to become part of COO eventually, and I would hope that we will have a clean CRD interface to that as opposed to creating configmaps in other namespaces - we'll see, though.

controllers/metricstorage_controller.go
csibbitt: Thanks for the confirmation, Juan. I've been experimenting with this a bit yesterday and today and I've come to the same conclusion.

controllers/metricstorage_controller.go
```diff
@@ -265,6 +268,96 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+                    plugin:
+                        kind: "PrometheusDatasource"
+                        spec:
+                            direct_url: "http://prometheus-operated.` + instance.Namespace + ".svc.cluster.local:9090\"",
+		}
+		return nil
+	})
+	if err != nil {
+		Log.Error(err, "Failed to update Console UI Datasource ConfigMap %s - operation: %s", datasourceCM.Name, string(op))
+		return ctrl.Result{}, err
```
csibbitt, line 334: I've just noticed that this is potentially a problem - I think that if the console-dashboards-plugin is not installed, then the "console-dashboards" namespace will not exist, and this will block the rest of the reconciliation. We'll need to test it, but I think the practical effect of deploying the dashboards without the console-dashboards-plugin is that the dashboard shows, but it connects to the openshift-monitoring prometheus instead of ours. (that's what happens if you don't install the datasource, anyways).

I'm wondering if I should log the error but not call return so that this doesn't interrupt the reconciliation?

This should not be a problem for the other resources. The PrometheusRule CRD will exist if we've successfully deployed a MonitoringStack, and the "openshift-config-managed" namespace will always exist.

jlarriba:
According to Kuttl:

`go/pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: failed to list *v1.PrometheusRule: prometheusrules.monitoring.rhobs is forbidden: User "system:serviceaccount:openstack-operators:telemetry-operator-controller-manager" cannot list resource "prometheusrules" in API group "monitoring.rhobs" at the cluster scope`

Permissions for prometheusrules.monitoring.rhobs are needed.

controllers/metricstorage_controller.go
jlarriba: While I agree that block the rest of the reconciliation because the plugin does not exist, I dont think silently connecting to openshift-monitoring prometheus in this case is a good user experience.

I would say that the "right" behaviour for when that situation happens is to log the error, do not interrupt reconciliation, fail the PrometheusRuleReadyCondition with an informative message and do not install the dashboards.

That would also include keep retrying reconciliations, but not every second, every 30 or 60 seconds for example. So when the user is informed by the Status about the problem, he can go on, install the plugin and the next reconciliation on the T-O will correctly install the dashboards.

WDYT?

csibbitt:
> According to [Kuttl](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_telemetry-operator/319/pull-ci-openstack-k8s-operators-telemetry-operator-main-telemetry-operator-build-deploy-kuttl/1763001236732252160/artifacts/telemetry-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-f745587ef185f9023b8c83c4d40a4e1c7478de462a91dc6280e5277545650bbe/namespaces/openstack-operators/pods/telemetry-operator-controller-manager-6b9b5d7f4c-zk2mt/logs/manager.log):
> 
> `go/pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: failed to list *v1.PrometheusRule: prometheusrules.monitoring.rhobs is forbidden: User "system:serviceaccount:openstack-operators:telemetry-operator-controller-manager" cannot list resource "prometheusrules" in API group "monitoring.rhobs" at the cluster scope`
> 
> Permissions for prometheusrules.monitoring.rhobs are needed.

Thanks for digging this out of the logs (and your link helped me figure out where to find them - not exactly obvious!)

I was wondering about this, but forgot about it when all the tests passed locally. I guess because I was running the operator as kubeadmin. 

Changes coming soon.

controllers/metricstorage_controller.go
csibbitt: Thanks for the suggestion!

I agree that silently connecting the the other prometheus is not a good outcome, and I think we should raise this as a bug against console UI. I like your idea of not installing the dashboards if the datasource fails to install, that seems like a great workaround!

For the other part, though, I'm wondering what other effects there might be to never having the PrometheusRuleReady condition satisfied? First, it's misleading because actually those rules might be installed and working. Second, if it never goes Ready, then (I think?) the MetricStorage will never go Ready, and I am not yet familiar enough to know what effects this will have on the user experience, or testing for that matter. I'm not against this if you think it will be okay, I'm just unsure at the moment.

It looks like my most recent CI kuttl tests failed for exactly this reason (no console-dashboards namespace, so no matching datasource object). If that's the current state, and we want to go further and not install dashboards without the datasource, then I think I will need to simply remove my new kuttl tests for now, pending delivery of our dependencies.

csibbitt:
I added the suggested logic of not installing dashboards if the datasource fails, and marking the associated PrometheusRule object NotReady. I also renamed the PrometheusRule-related code to more closely associate it with dashboards. In the future, we will easily distinguish these rules from other rules for another purpose like alerts, or OTel compatibility.

All tests passed locally at this point (with console-dashboards-plugin installed)

Finally, I removed some new kuttl tests will not be able to pass until we can bring in console-dashboards-plugin as a dependency.

controllers/metricstorage_controller.go
jlarriba: There are no effects on not having the PrometheusRuleReady condition satisfied. What it can never happen is that the rules are installed and working and the PrometheusRuleReady condition is not ready, then we are marking a condition that it is not. 

Maybe I am confused here and not understanding the problem. If the rules installed and working does not depend on the "console-dashboard" namespace existence, then we are not using the correct condition. We might  need a new condition: one that reports the wrong status because there is some "stuff" that we couldn't do due to the namespace not existing (the user did not created the pre-requisites).

Leaving one of the status of the operator as not ready is not an issue for the rest of the MetricStorage, which would get deployed and usable even the operator is not ready. Of course, if the dashboards never go Ready, the MetricStorage never goes ready, but this is implicit behaviour we want to have and that's why we have decided to mix MetricStorage and Dashboards under the same operator "api". If we would like for them to be different actions, we would have created different "api". However, we might want to allow the user to enable/disable dashboarding exactly the same way we are allowing that for alertmanager, but I don't see this as a hard requirement.

Take in account that when an operator is not ready, from a user experience perspective, indicates that there is an issue that must be solved, so I dont really see why the user would leave the operator as not ready forever: he should be taking corrective action (installing prerequisite). If we really think that he would like to have a MetricStorage without a Dashboard, then the option of disabling it would make sense. I think this is a very strange use case.

csibbitt: AFAICT, if the the PrometheusRuleReady (or some other API we invent) never goes Ready, then the Telemetry object never goes Ready, and then the autoscaling tests will fail because they wait for it to be Ready. (I have a slack thread with Emma about this).

No matter which API we choose or what options the user may or may not want, I don't think it is possible for us to successfully test functionality that depends on console-dashboards-plugin until it's actually available.

The solution is either to not report errors (and let things go "Ready" even though the dashboards aren't), or we simply block this work from merging until the console-dashboards-plugin is actually available.

controllers/ceilometer_controller.go
```diff
@@ -565,11 +565,11 @@ func (r *CeilometerReconciler) generateComputeServiceConfig(
 	ceilometerPasswordSecret, _, _ := secret.GetSecret(ctx, h, instance.Spec.Secret, instance.Namespace)
 
 	templateParameters := map[string]interface{}{
-		"KeystoneInternalURL":           keystoneInternalURL,
```
csibbitt, line 568: This whitespace change occurs every time I run `make run`, and I got tired of reverting it

csibbitt:
I added the extra status indicators as we discussed. Still, this is not going to pass the autoscaling tests until the console-dashboards-plugin is available (because the datasource and dashboards will never go Ready). If we're going to wait for that, I suppose I might as well just add the kuttl tests back, too? WDYT, @jlarriba ?

controllers/ceilometer_controller.go
jlarriba: This is gofmt formatting the code, do not correct it and let it be like this.

jlarriba:
As it has a hard requirement on something external, my recommendation would be to make this optional by adding an extra "enabled" switch. That way we can leave the code there, disable it, pass all the tests and when the console-dashboards-plugin is available in the cluster, we could just instruct the user to switch the "enabled" to on.

This will protect us if we reach GA without the console-dashboards-plugin readily available but will allow us to ship the code.

csibbitt:
Replaced by https://github.com/openstack-k8s-operators/telemetry-operator/pull/328  because I've been asked to delete my branches in this repo

PR #320, opened 2024-02-28
Add TLS support for Prometheus and sg-core

vyzigold:
This PR adds TLS support for Prometheus and sg-core.

Prometheus web UI is put behind TLS with certificates generated by openstack-operator. ServiceMonitor and ScrapeConfig are configured to use CA certs to correctly access metrics from sg-core and NodeExporter with TLS when it's enabled.

A new proxy container is put to the ceilometer-0 pod in front of sg-core. This container runs httpd to terminate the TLS connections, because sg-core doesn't support TLS natively.

Alertmanager is left without TLS, because COO limitations wouldn't allow prometheus to reach alertmanager if it's secured with TLS. There is a feature request created for COO, which should enable us to add TLS for alertmanager in the future. See rh-jira: OSPRH-5177 and COO-44.

What's missing:
- [ ]  openstack-operator PR for creating the routes and certificates
- [ ] Go through the sg-core half of the PR, add ifs, remove hardcoded values, ...
- [ ] Test in a clean environment
- [ ] Test that prometheus can reach NodeExporter on the dataplane when it uses TLS
- [ ] kuttl-tests

What's working with TLS enabled:
- [x] User can access prometheus with https when using the combined-ca-bundle
- [x] User can access sg-core with https when using the combined-ca-bundle
- [x] Prometheus can access sg-core
- [ ] Prometheus can access NodeExporter - this should work, but is untested
- [x] Aodh can access prometheus. Works with combined-ca-bundle and with this patch: https://review.opendev.org/c/openstack/python-observabilityclient/+/910229


vyzigold:
There is a parallel PR https://github.com/openstack-k8s-operators/telemetry-operator/pull/310 on adding TLS support to aodh and ceilometer, which also changes a lot of the same files as the sg-core related changes here. The overlap is a little bigger than I initially anticipated, but that PR looks on a good way to be merge-able soon. I'll wait with refactoring the sg-core related changes until it's merged and I can refactor on top of it.

api/v1beta1/ceilometer_types.go
```diff
@@ -37,6 +38,8 @@ const (
 	// NodeExporterContainerImage - default fall-back image for node_exporter
 	// NodeExporterContainerImage = "registry.redhat.io/openshift4/ose-prometheus-node-exporter:v4.13"
 	NodeExporterContainerImage = "quay.io/prometheus/node-exporter:v1.5.0"
+	// ProxyContainerImage - default fall-back image for proxy container
+	ProxyContainerImage     = "quay.io/podified-antelope-centos9/openstack-swift-proxy-server:current-podified"
```
vyzigold, line 42: The openstack-aodh-api image should work here as well and at least it's telemetry-related. I'll change it to that unless there are other suggestions.

api/v1beta1/ceilometer_types.go
```diff
@@ -90,6 +93,11 @@ type CeilometerSpec struct {
 
 	// +kubebuilder:validation:Required
 	NodeExporterImage string `json:"nodeExporterImage"`
+
+	// +kubebuilder:validation:Required
+	ProxyImage string `json:"proxyImage"`
+
+	TLS tls.API `json:"tls,omitempty"`
```
vyzigold, line 100: There are 3 containers in the ceilometers pod communicating with the world outside.

- Ceilometer containers communicate with the rest of the openstack services and with rabbit, but they don't have an api (they don't receive connections from outside the pod).
- sg-core container (after this PR it'll be the proxy container) receives connections from prometheus, but I don't see a reason for the user to access it or for a route to be created. sg-core doesn't initiate connections to outside of the pod.

What would be the correct type here? tls.SimpleService? And mount the ca cert to ceilometer containers and the key with cert to sg-core?

api/v1beta1/conditions.go
```diff
@@ -156,6 +162,24 @@ const (
...
+
+	//
+	// PrometheusServiceReady condition messages
+	//
+	// PrometheusServiceReadyInitMessage
+	PrometheusServiceReadyInitMessage = "Prometheus service not watched"
+
+	// PrometheusServiceUnableToWatchMessage
+	PrometheusServiceUnableToWatchMessage = "Error occured when trying to watch %s"
+
```
vyzigold, line 182: TODO: delete this. We don't need to "dynamically" add watches for services.

api/v1beta1/metricstorage_types.go
```diff
@@ -95,6 +96,21 @@ type MetricStorageSpec struct {
 	// +kubebuilder:validation:Optional
 	// +nullable
 	CustomMonitoringStack *obov1.MonitoringStackSpec `json:"customMonitoringStack,omitempty"`
+
+	// TODO: Is this the correct type?
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// TLS - Parameters related to the TLS
+	PrometheusTLS tls.SimpleService `json:"prometheusTls,omitempty"`
```
vyzigold, line 104: There isn't any internal endpoint for Prometheus. COO creates only one service and after a future PR the openstack-operator will create a route for that. Is the SimpleService the type meant for such situations?

controllers/metricstorage_controller.go
```diff
@@ -57,6 +62,19 @@ import (
 	obov1 "github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
 )
 
+// fields to index to reconcile when change
+const (
+	caBundleSecretNameField = ".spec.tls.simpleService.caBundleSecretName"
+	tlsField                = ".spec.tls.simpleService.secretName"
```
vyzigold, line 68: spec.prometheusTls.* instead of these two lines

controllers/metricstorage_controller.go
```diff
@@ -247,6 +268,54 @@ func (r *MetricStorageReconciler) reconcileNormal(
 	if op != controllerutil.OperationResultNone {
 		Log.Info(fmt.Sprintf("MonitoringStack %s successfully changed - operation: %s", monitoringStack.Name, string(op)))
 	}
+
+	if instance.Spec.PrometheusTLS.Enabled() {
+		// TODO: When switching from TLS enabled to TLS disabled, we need to "unpatch" prometheus
+		//       How to detect that? How to "unpatch"? I guess I could just delete the CR, but we don't want to do that every time as
+		//       the reconciliation of prometheus after deleting the CR takes long (like a minute)
```
vyzigold, line 275: I guess I could just add an else branch here. I'd get the prometheus CR and check if it has tlsConfig. If it does and tls is supposed to be disabled, I'll delete the prometheus CR and let it be recreated without the patch by COO. I'm open to other ideas.

go.mod
```diff
@@ -98,3 +98,5 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202304141430
 // needed to to cert-manager v1.11.4 see https://github.com/cert-manager/cert-manager/blob/v1.11.4/go.mod#L263C1-L264C104
 // remove this once we bump to cert-manager v1.12.x
 replace github.com/Venafi/vcert/v4 => github.com/jetstack/vcert/v4 v4.9.6-0.20230519122548-219f317ae107 //allow-merging
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => ../lib-common/modules/common
```
vyzigold, line 102: TODO: this should go away after a rebase. If there are failures, then update lib-common.

pkg/ceilometer/service.go
```diff
@@ -34,8 +35,11 @@ func Service(
 ) (*corev1.Service, controllerutil.OperationResult, error) {
 	service := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ServiceName,
+			Name:      "ceilometer-internal",
```
vyzigold, line 38: This shouldn't be hardcoded

pkg/metricstorage/prometheus_tls_patch.go
```diff
@@ -0,0 +1,69 @@
...
+		Spec: monv1.PrometheusSpec{
+			CommonPrometheusFields: monv1.CommonPrometheusFields{
+				Web: &monv1.PrometheusWebSpec{
+					WebConfigFileFields: monv1.WebConfigFileFields{
+						TLSConfig: &monv1.WebTLSConfig{
+							KeySecret: corev1.SecretKeySelector{
+								LocalObjectReference: corev1.LocalObjectReference{
+									Name: *instance.Spec.PrometheusTLS.SecretName,
+								},
+								Key: "tls.key",
```
vyzigold, line 48: Can the key in the cert secret stay hardcoded like this or do we have a const somewhere?

pkg/metricstorage/prometheus_tls_patch.go
```diff
@@ -0,0 +1,69 @@
...
+							},
+							Cert: monv1.SecretOrConfigMap{
+								Secret: &corev1.SecretKeySelector{
+									LocalObjectReference: corev1.LocalObjectReference{
+										Name: *instance.Spec.PrometheusTLS.SecretName,
+									},
+									Key: "tls.crt",
```
vyzigold, line 55: Can the key in the cert secret stay hardcoded like this or do we have a const somewhere?

pkg/metricstorage/prometheus_tls_patch.go
```diff
@@ -0,0 +1,69 @@
...
+								},
+							},
+						},
+					},
+				},
+				Secrets: []string{
+					// TODO: Is this the correct ca secret name for reaching sg-core (ceilometer)?
+					instance.Spec.PrometheusTLS.CaBundleSecretName,
```
vyzigold, line 63: I know, there is the internal and public rootCAs. I'd consider prometheus as a public service, but it needs to reach sg-core, which I'd consider internal. I know Prometheus was able to reach sg-core when I configured this value as "combined-ca-bundle". I believe by default openstack-operator is setting these values if TLS is enabled for the control-plane.

Can I expect that the prometheus will be able to reach sg-core in that case?
Can I expect that the prometheus will be able to reach NodeExporter, which is running on the dataplane?

pkg/metricstorage/scrape_config.go
```diff
@@ -42,16 +42,32 @@ func ScrapeConfig(
...
+
+	if instance.Spec.PrometheusTLS.Enabled() {
+		spec["scheme"] = "HTTPS"
+		spec["tlsConfig"] = map[string]interface{}{
+			"ca": map[string]interface{}{
+				// TODO: Is this the correct secret for reaching node-exporter on the dataplane?
+				"secret": map[string]interface{}{
+					"name": instance.Spec.PrometheusTLS.CaBundleSecretName,
+					// TODO: What about this value? Should that be hardcoded?
+					"key": "ca-bundle.pem",
```
vyzigold, line 62: This is the key for a value in the CaBundleSecretName. Should I leave it hardcoded like this or do we have a const somewhere?

pkg/metricstorage/service_monitor.go
```diff
@@ -82,5 +85,15 @@ func ServiceMonitor(
 			},
 		},
 	}
+	if instance.Spec.PrometheusTLS.Enabled() {
+		serviceMonitor.Spec.Endpoints[0].Scheme = "https"
+		serviceMonitor.Spec.Endpoints[0].TLSConfig = &monv1.TLSConfig{
+			// TODO: Is this the correct way to get the CA bundle secret name? What about the filename inside the secret?
+			CAFile: fmt.Sprintf("/etc/prometheus/secrets/%s/internal-ca-bundle.pem", instance.Spec.PrometheusTLS.CaBundleSecretName),
```
vyzigold, line 92: Prometheus is using the internal-ca-bundle.pem from the CaBundleSecretName. Should I leave it hardcoded like this or do we have a const somewhere?

stuggi: I am not sure I understand how the ca bundle gets mounted into the prometheus pod

controllers/metricstorage_controller.go
```diff
@@ -375,6 +444,55 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.TLSInputErrorMessage,
+				err.Error()))
+			return ctrlResult, err
+		} else if (ctrlResult != ctrl.Result{}) {
+			return ctrlResult, nil
+		}
+		// TODO: What to do with hash? Do we needed for metricstorage? We don't manage secrets / cms. Prometheus CR is currently patched on every reconcile.
+		fmt.Println(hash)
```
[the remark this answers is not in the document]
stuggi: how would you trigger a restart of prometheus when the certs get updated (content of the secret)

controllers/metricstorage_controller.go
stuggi: I don't think that will work, but I haven't tested. to add fields to the indexer I think you must name them, but we have it as a task for later to see if we can improve indexing additional fields.

vyzigold: I made the comment mainly for me to not forget about this and I may have not explained correctly what I want to do. You mean, that if I change it to the following, it won't work?
```
        caBundleSecretNameField = ".spec.prometheusTls.caBundleSecretName"
	tlsField                                    = ".spec.prometheusTls.secretName"
```

controllers/metricstorage_controller.go
vyzigold: The Prometheus CR is owned and created by COO, so patching it out and just deleting it will have the same result. The only difference is, that deleting it will make COO reconcile, while patching it out wouldn't be noticed by COO. Prometheus will restart either way. I see 2 advantages in just deleting the CR. It's easier and I can be sure, than I didn't forget to patch out something in the CR.

controllers/metricstorage_controller.go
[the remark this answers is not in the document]
vyzigold: Thanks for the explanation. I guess this will need some investigation on how to handle this for prometheus. It's possible that the Prometheus operator would watch these secrets and do the restarts or maybe I could somehow include them as env vars.

controllers/metricstorage_controller.go
stuggi: ah ok, that should be fine. I understood you wanted to use some `.*` wildcard

controllers/metricstorage_controller.go
stuggi: I am not exactly sure what COO is but if it is owned by a technical instance where we can be sure it re-creates it, it could be fine. In general we just should be careful to delete an object which I do not own. we also need to check if there could be a finalizer which blocks deleting it?

pkg/metricstorage/service_monitor.go
vyzigold: On this line: https://github.com/openstack-k8s-operators/telemetry-operator/blob/c5a5800c1a34a56de98b6e160dd772e037f3155b/pkg/metricstorage/prometheus_tls_patch.go#L63 Prometheus operator mounts the secret to `/etc/prometheus/setrecs/<secret_name>` . Service monitors can then use this

controllers/metricstorage_controller.go
vyzigold: The hierarchy of monitoring resources is a little confusing. There are 3 operators that play some part in monitoring. COO = cluster-observability-operator.

telemetry-operator creates the MonitoringStack resource - this is owned by telemetry-operator. Based on that COO creates the Prometheus and Alertmanager resources - these are owned by COO, but we can use server-side apply to patch some parts of them. Based on these the prometheus-operator creates statefulsets, services, ...

I'll check if there are finalizers, but I don't think so.

Deydra71:
> There is a parallel PR #310 on adding TLS support to aodh and ceilometer, which also changes a lot of the same files as the sg-core related changes here. The overlap is a little bigger than I initially anticipated, but that PR looks on a good way to be merge-able soon. I'll wait with refactoring the sg-core related changes until it's merged and I can refactor on top of it.

I'm now confused about the changes both of us do to the Ceilometer part. If I understand correctly, we want to use the httpd running in a `ceilometer-0` container in order to terminate TLS for sg-core, correct? Sg-core is an API service, and I see the conflict in this PR and #310 where we both implement different TLS types in the Ceilometer service. If Ceilometer will newly run httpd, I should probably drop my changes to Ceilometer in favor of your patch to implement everything Ceilometer needs for successful sg-core TLS connection.

What do you think? I hope my thoughts are understandable.

vyzigold:
@Deydra71 The split between ceilometer and sg-core TLS work isn't as clean as I initially thought (I'm sorry for that), because they share the same pod, controller, etc. Ceilometer needs just the CA certificates mounted, while sg-core needs it's own certificate, key and a proxy in front of it. I'm not sure how far you're with your ceilometer work on https://github.com/openstack-k8s-operators/telemetry-operator/pull/310 . If you're close to being finished, I can rebase this PR once 310 is merged and deal with the changes needed for sg-core. If there is some significant additional work required for ceilometer in 310, which would need to be changed by this PR, then feel free to drop the changes. It wouldn't make sense to work on something, which would get rewritten after a few days. In my understanding the biggest conflict so far is because of the different TLS type right?

Deydra71:
> @Deydra71 The split between ceilometer and sg-core TLS work isn't as clean as I initially thought (I'm sorry for that), because they share the same pod, controller, etc. Ceilometer needs just the CA certificates mounted, while sg-core needs it's own certificate, key and a proxy in front of it. I'm not sure how far you're with your ceilometer work on #310 . If you're close to being finished, I can rebase this PR once 310 is merged and deal with the changes needed for sg-core. If there is some significant additional work required for ceilometer in 310, which would need to be changed by this PR, then feel free to drop the changes. It wouldn't make sense to work on something, which would get rewritten after a few days. In my understanding the biggest conflict so far is because of the different TLS type right?

Since yesterday, I've been trying to resolve the Ceilometer pod crashing because the internal Keystone endpoint is unreachable (haven't found out the reason so far). But yes, my changes conflict with the work you need to do for sg-core. I'll drop my changes to Ceilometer for you to freely implement what you need for sg-core, because you have a better understanding of what needs to be done.

The TLS for Aodh is working properlym, I just need to finish the kuttl tests. So without the ceilometer, my patch is basically ready.

Thanks!

controllers/metricstorage_controller.go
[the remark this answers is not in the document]
vyzigold: I just tested this and it looks like the prometheus-operator takes care of updating everything when the certs change.
Testing procedure I used:
- `oc edit cert-metric-storage-prometheus-svc` and do a random change to the tls.crt. This seems to make cert-manager to regenerate the secret
- `oc extract secret/cert-metric-storage-prometheus-svc --keys=tls.crt --to=- | md5sum` to verify the certificate changed (it did)
- After about 30s: `oc rsh openstackclient openssl s_client -connect metric-storage-prometheus.openstack.svc:9090 < /dev/null 2> /dev/null | openssl x509 -inform pem | md5sum` this returned the same hash as the command above.

This was without telemetry-operator running, so the update definitely came from the prometheus-operator. I'll just discard the returned hashes, as it looks like they aren't needed

controllers/metricstorage_controller.go
vyzigold: The approach with deleting Prometheus CR worked.

vyzigold:
I tested the code and I've confirmed:
- It's able to correctly switch from TLS enabled to disabled and back
- Prometheus, aodh, sg-core and node-exporter endpoints are using https.
- Prometheus is able to reach node-exporter and sg-core
- When using the correct aodh images and the patch mentioned in PR description, I was able to create alarms and they got successfully evaluated, which confirms, that aodh can reach Prometheus.
- None of the telemetry containers have logs suggesting TLS issues

It'd be a good idea to do our autoscaling test we performed a while back again, but with TLS enabled once the rest of the services are finished with implementing TLS (even on the dataplane).

Once the tests pass, I'll squash the commits and this will be ready for review.

vyzigold:
memcached error:
```
  Warning  Failed          25m (x6 over 27m)     kubelet            Error: ImagePullBackOff
  Normal   Pulling         25m (x4 over 27m)     kubelet            Pulling image "quay.io/podified-antelope-centos9/openstack-memcached@sha256:8347d7e59d00117771cdc68c39c2ccebcfb6dbd197689bd2231a1e9fd8fb44c6"
  Warning  Failed          25m (x4 over 27m)     kubelet            Failed to pull image "quay.io/podified-antelope-centos9/openstack-memcached@sha256:8347d7e59d00117771cdc68c39c2ccebcfb6dbd197689bd2231a1e9fd8fb44c6": rpc error: code = Unknown desc = reading manifest sha256:8347d7e59d00117771cdc68c39c2ccebcfb6dbd197689bd2231a1e9fd8fb44c6 in quay.io/podified-antelope-centos9/openstack-memcached: manifest unknown
  Warning  Failed          25m (x4 over 27m)     kubelet            Error: ErrImagePull
  Normal   BackOff         118s (x109 over 27m)  kubelet            Back-off pulling image "quay.io/podified-antelope-centos9/openstack-memcached@sha256:8347d7e59d00117771cdc68c39c2ccebcfb6dbd197689bd2231a1e9fd8fb44c6"
```

vyzigold:
Tests are passing. I think this is ready for reviews.

vyzigold:
recheck

api/v1beta1/metricstorage_types.go
```diff
@@ -100,11 +101,26 @@ type MetricStorageSpec struct {
 	// +kubebuilder:validation:Optional
 	// +nullable
 	CustomMonitoringStack *obov1.MonitoringStackSpec `json:"customMonitoringStack,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// TLS - Parameters related to the TLS
+	PrometheusTLS tls.SimpleService `json:"prometheusTls,omitempty"`
+
+	// TODO: Implement TLS for alertmanager Web UI
```
stuggi, line 110: is this something we have to document as a current limitation for tlse  to show that this is known and not "forgotten"?

ci/vars-autoscaling-tempest.yml
```diff
@@ -12,7 +12,7 @@ cifmw_tempest_tempestconf_config:
       service_available.ceilometer true
       service_available.sg_core true
       service_available.aodh true
-      telemetry.sg_core_service_url "ceilometer.openstack.svc.cluster.local:3000"
+      telemetry.sg_core_service_url "ceilometer-internal.openstack.svc.cluster.local:3000"
```
stuggi, line 15: could it be that this will run in a different namespace then  openstack?

templates/ceilometercentral/config/ceilometer.conf
```diff
@@ -16,6 +16,9 @@ username=ceilometer
 user_domain_name=Default
 password={{ .CeilometerPassword }}
 interface = internalURL
+{{- if .TLS }}
+cafile = {{ .CAFile }}
```
stuggi, line 20: is this something we have to set as we mount the bundle to the system wide CA bundle? In other services we removed that as long it does not is a special one in a special path.

controllers/ceilometer_controller.go
```diff
@@ -513,7 +584,19 @@ func (r *CeilometerReconciler) generateServiceConfig(
...
 		"CeilometerPassword":  string(ceilometerPasswordSecret.Data["CeilometerPassword"]),
+		"TLS":                 false, // Default to false. Change to true later if TLS enabled
+	}
+
+	// create httpd  vhost template parameters
+	endptConfig := map[string]interface{}{}
+	endptConfig["ServerName"] = fmt.Sprintf("%s-internal.%s.svc", ceilometer.ServiceName, instance.Namespace)
+	if instance.Spec.TLS.Enabled() {
+		templateParameters["TLS"] = true
+		templateParameters["CAFile"] = tls.DownstreamTLSCABundlePath
```
stuggi, line 595: it won't hurt, but do we need it as its the sytem wide bundle?

controllers/metricstorage_controller.go
```diff
@@ -253,6 +274,71 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+	} else if instance.Status.PrometheusTLSPatched {
+		// Delete the prometheus CR, so it can be automatically restored without the TLS patch
+		prometheus := monv1.Prometheus{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: instance.Namespace,
+				Name:      instance.Name,
+			},
+		}
+		err = r.Client.Delete(context.Background(), &prometheus)
+		if err != nil {
```
stuggi, line 312: maybe we want to ignore NotFound?

controllers/metricstorage_controller.go
```diff
@@ -253,6 +274,71 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+			instance.Status.Conditions.MarkFalse(telemetryv1.PrometheusReadyCondition,
+				condition.Reason("Can't remove TLS configuration from Prometheus"),
```
stuggi, line 314: is this reason correct? should it be something like the error message bellow?

controllers/telemetry_controller.go
```diff
@@ -397,6 +397,8 @@ func reconcileMetricStorage(ctx context.Context, instance *telemetryv1.Telemetry
 			metricStorageInstance.Spec.MonitoringStack = &telemetryv1.MonitoringStack{}
 			instance.Spec.MetricStorage.MetricStorageSpec.MonitoringStack.DeepCopyInto(metricStorageInstance.Spec.MonitoringStack)
 		}
+		instance.Spec.MetricStorage.MetricStorageSpec.PrometheusTLS.DeepCopyInto(&metricStorageInstance.Spec.PrometheusTLS)
+		//instance.Spec.MetricStorage.MetricStorageSpec.AlertmanagerTLS.DeepCopyInto(&metricStorageInstance.Spec.AlertmanagerTLS)
```
stuggi, line 401: remove if not required?

pkg/metricstorage/alertmanager_service_patch.go
```diff
@@ -0,0 +1,49 @@
...
+) corev1.Service {
+	svc := corev1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-alertmanager", instance.Name),
+			Namespace: instance.Namespace,
+			Annotations: map[string]string{
```
stuggi, line 41: do they get merged, or do those need to be merged  with any existing annotations? we had some trouble a while back where the metallb operator with a new version started to add annotation on a successful ip reservation on the service. because we expected to be the only one adding annotations at that time there was an update race.

pkg/metricstorage/prometheus_service_patch.go
```diff
@@ -0,0 +1,49 @@
...
+) corev1.Service {
+	svc := corev1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-prometheus", instance.Name),
+			Namespace: instance.Namespace,
+			Annotations: map[string]string{
```
stuggi, line 41: same as above, do they get merged, or do those need to be merged with any existing annotations?

ci/vars-autoscaling-tempest.yml
vyzigold: I don't know. As far as I know tempest runs only in openstack namespace. I'm not sure if there is some variable we can use. What do you think @elfiesmelfie

controllers/ceilometer_controller.go
vyzigold: I'm not sure what's the question. You mean I should try using the `tls.UpstreamTLSCABundlePath` instead? Ceilometer is accessing APIs of a lot of the other openstack services, but it doesn't connect to anything else. So as long as it lets ceilometer access the internal endpoints, then we can use some other CA path.

controllers/metricstorage_controller.go
vyzigold: The code is deleting prometheus to remove the TLS configuration. My thoughts when I came up with the reason and log message was, that the user needs to know that we're trying to remove the TLS configuration. The user doesn't need to know, that we're deleting it. And I for some reason wanted to be a little more specific about what's the operator actually trying to do in the log message.

Looking at it now, maybe it would make a little more sense to unify the messages. Would "Can't delete old Prometheus CR to remove TLS configuration" sound better for both of them?

controllers/telemetry_controller.go
vyzigold: This line would be used to add TLS to alertmanager. I left it there as a reminder for when we implement it. I guess I could remove it or add a "TODO" comment there to show that it's something to be used in the future.

pkg/metricstorage/alertmanager_service_patch.go
vyzigold: So right now there aren't any annotations from COO or prometheus-operator, so the code for now works like this. I guess it's possible, that this will change in the future. How did you work around that with metallb?

ci/vars-autoscaling-tempest.yml
stuggi: yes, I think tempest right now is only running in that namespace. probably something we need to keep in mind if this changes.

templates/ceilometercentral/config/ceilometer.conf
stuggi: ok, seems it behaves different then the other services

controllers/ceilometer_controller.go
stuggi: was related to https://github.com/openstack-k8s-operators/telemetry-operator/pull/320#discussion_r1523485341 . if the services won't use the system wide bundle per default of not configured in the cfg. but its not a problem

controllers/metricstorage_controller.go
stuggi: yes, that sounds good to me

controllers/telemetry_controller.go
stuggi: yes, I'd either remove it, or add a TODO comment that someone like me knows that it is intentionally in

pkg/metricstorage/alertmanager_service_patch.go
stuggi: we solved it, merging the annotations with current existing ones https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service.go#L311 .

ci/vars-autoscaling-tempest.yml
elfiesmelfie: I don't think it makes sense to ever change where tempest runs. The test-operator controls this though.

Why does this URL need to change here? Is it a symptom of some kind of incorrect configuration that the public URL doesn't work?

vyzigold: I just aligned the ceilometer service name to the rest of the services. All the other services end either with "-internal" or "-public". In my opinion ceilometer (sg-core) behaves as an internal endpoint. We don't expose it for the user.

pkg/metricstorage/alertmanager_service_patch.go
vyzigold: I just tested what currently happens if there are annotations and the operator patches the service. The procedure I did:
- stop telemetry-operator
- oc edit the prometheus and alertmanager services. I deleted all the annotations added by telemetry-operator and I added a new annotation instead.
- start telemetry-operator again

When I oc get the services, I see it merged the annotations (I can see the annotation I created as well as the annotations created by telemetry-operator). So this should be alright.

vyzigold:
recheck

Falied while waiting for oscp to complete

jlarriba:
/lgtm

For the next time, we should have divided this PR into multiple PRs and do an incremental approach to TLS enablement, component by component. That's usually always better than a huge big bang.

vyzigold:
I believe I encountered the current error. I proposed a fix to openstack-operator, which should fix it: https://github.com/openstack-k8s-operators/openstack-operator/pull/711

vyzigold:
> /lgtm
> 
> For the next time, we should have divided this PR into multiple PRs and do an incremental approach to TLS enablement, component by component. That's usually better than a huge big bang.

I totally agree. At the start I thought it'll be smaller, but it just kept growing and growing.

vyzigold:
recheck
Fixes for the issue are now merged.

vyzigold:
recheck
Logs from ceilometer central agent:
```
2024-03-17 09:16:16.438 14 ERROR ceilometer.polling.manager [-] Skipping images, keystone issue: internalURL endpoint for image service not found: keystoneauth1.exceptions.catalog.EndpointNotFound: internalURL endpoint for image service not found
```
We may have some timing issue.

jlarriba:
> recheck Logs from ceilometer central agent:
> 
> ```
> 2024-03-17 09:16:16.438 14 ERROR ceilometer.polling.manager [-] Skipping images, keystone issue: internalURL endpoint for image service not found: keystoneauth1.exceptions.catalog.EndpointNotFound: internalURL endpoint for image service not found
> ```
> 
> We may have some timing issue.

Yes, I have found this many times, but it fixes itself when Keystone configurtion is sorted out. It is the result of a race condition caused by running all the installation at the same time. This is an established pattern on Kubernetes, and as long as services are configured to support it (which seems to be the case), it should not cause an issue.

vyzigold:
/retest
I think the issue should be resolved now.

vyzigold:
Now there is a similar error with heat image:
```
  Normal   Scheduled       22m                   default-scheduler  Successfully assigned telemetry-kuttl-tests/heat-db-sync-tn5tk to oko-08-qszq7-master-0
  Normal   AddedInterface  22m                   multus             Add eth0 [10.128.0.145/23] from ovn-kubernetes
  Normal   Pulling         20m (x4 over 22m)     kubelet            Pulling image "quay.io/podified-antelope-centos9/openstack-heat-api@sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda"
  Warning  Failed          20m (x4 over 22m)     kubelet            Failed to pull image "quay.io/podified-antelope-centos9/openstack-heat-api@sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda": rpc error: code = Unknown desc = reading manifest sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda in quay.io/podified-antelope-centos9/openstack-heat-api: manifest unknown
  Warning  Failed          20m (x4 over 22m)     kubelet            Error: ErrImagePull
  Warning  Failed          20m (x6 over 22m)     kubelet            Error: ImagePullBackOff
  Normal   BackOff         2m19s (x85 over 22m)  kubelet            Back-off pulling image "quay.io/podified-antelope-centos9/openstack-heat-api@sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda"

```

karelyatin:
/test telemetry-operator-build-deploy-kuttl

latest heat-operator-bundle refreshed

PR #327, opened 2024-03-06
Add missing aodh httpd config

vyzigold:
There was configuration missing to enable TLS on aodh api endpoints. The httpd and wsgi files are taken from the cinder-operator. I was able to verify this works
by executing curl from openstackclient pod against the https://aodh-internal.openstack.svc and
https://aodh-public.openstack.svc endpoints.
The "openstack alarm list" command started working as well after this change when TLS was enabled in the control plane.

I also renamed the field index variables in preparation for https://github.com/openstack-k8s-operators/telemetry-operator/pull/320. The reason is, that all of the controllers are in the same package. Each controller needs these variables defined slightly different, so we need a unique name for the variables across each controller.


Deydra71:
/lgtm

Thanks a lot for the catch and fix!

Deydra71 (approved):

PR #328, opened 2024-03-06
Adding Console UI Dashboards automation

csibbitt:
- This depends on the console-dashboards-plugin being installed w/ Helm as described here: https://github.com/openshift/console-dashboards-plugin?tab=readme-ov-file#deployment-on-cluster
- Creates Prometheus recording rules (equivalent to ones used in OCP) to drive the dashboards
- Creates a console UI datasource that points to our prometheus (See: https://github.com/openshift/console-dashboards-plugin/blob/main/docs/add-datasource.md)
- Creates Dashboards (See: https://github.com/openshift/console-dashboards-plugin/blob/main/docs/prometheus-dashboard-example.yaml)


jlarriba:
Re-posting my latest comment in the previous PR as it is still valid:

My recommendation would be to make this optional by adding an extra "enabled" switch inside the monitoringStack yaml that is able to enable or disable the dashboard creation. That way we can leave the code here, pass all the tests and when the console-dashboards-plugin is available in the cluster, we could just instruct the user to switch the "enabled" to on.

This will protect us if we reach GA without the console-dashboards-plugin readily available but will allow us to actually ship the code.

csibbitt:
> Re-posting my latest comment in the previous PR as it is still valid:
> 
> My recommendation would be to make this optional by adding an extra "enabled" switch inside the monitoringStack yaml that is able to enable or disable the dashboard creation. That way we can leave the code here, pass all the tests and when the console-dashboards-plugin is available in the cluster, we could just instruct the user to switch the "enabled" to on.
> 
> This will protect us if we reach GA without the console-dashboards-plugin readily available but will allow us to actually ship the code.

Okay, that makes sense, thanks. I'll add this ASAP

csibbitt:
Kuttl tests pass locally, but failed with these errors:

```
case.go:364: failed in step 1-deploy
    case.go:366: Internal error occurred: failed calling webhook "mtelemetry.kb.io": failed to call webhook: Post "[https://telemetry-operator-controller-manager-service.openstack-operators.svc:443/mutate-telemetry-openstack-org-v1beta1-telemetry?timeout=10s](https://telemetry-operator-controller-manager-service.openstack-operators.svc/mutate-telemetry-openstack-org-v1beta1-telemetry?timeout=10s)": no endpoints available for service "telemetry-operator-controller-manager-service"
```
and
```
case.go:364: failed in step 1-deploy
    case.go:366: Internal error occurred: failed calling webhook "mmetricstorage.kb.io": failed to call webhook: Post "[https://telemetry-operator-controller-manager-service.openstack-operators.svc:443/mutate-telemetry-openstack-org-v1beta1-metricstorage?timeout=10s](https://telemetry-operator-controller-manager-service.openstack-operators.svc/mutate-telemetry-openstack-org-v1beta1-metricstorage?timeout=10s)": no endpoints available for service "telemetry-operator-controller-manager-service"
```

I also see 
```
error: unable to upgrade connection: container container-00 not found in pod oko-09-7n5pz-master-2-debug-j9vln_openstack
make: *** [Makefile:559: crc_storage] Error 1
```
but honestly I'm not sure if this is fatal or just part of the "normal" raft of errors in the output?

Given that the content provider test failed, there appear to be deployment errors, and all my manual and automated tests are passing locally, I'm going to assume the problem is elsewhere and ask CI to retry everything.

csibbitt:
recheck

PR #337, opened 2024-03-12
Add kube-state-metrics service

paramite:
We need this service to monitor state of pods and their containers.


controllers/ceilometer_controller.go
```diff
@@ -452,7 +472,178 @@ func (r *CeilometerReconciler) reconcileNormal(ctx context.Context, instance *te
...
+	// create the service
+	serviceLabels := map[string]string{
+		common.AppSelector: availability.KSMServiceName,
+	}
+
+	deployDef, err := availability.KSMDeployment(instance, serviceLabels)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	deploy := deployment.NewDeployment(
```
vyzigold, line 547: A while ago a lot of the services including us moved from deployments to statefulsets. Is there a reason to use a deployment instead of statefulset? I think it would be better to keep it consistent with the rest of the telemetry if possible.

controllers/metricstorage_controller.go
```diff
@@ -298,6 +299,29 @@ func (r *MetricStorageReconciler) reconcileNormal(
 	if op != controllerutil.OperationResultNone {
 		Log.Info(fmt.Sprintf("Ceilometer ServiceMonitor %s successfully changed - operation: %s", ceilometerMonitor.Name, string(op)))
 	}
+
+	ksmMonitor := &monv1.ServiceMonitor{
```
vyzigold, line 303: I'm not sure about this section of code. It looks to me like this may be overwriting the ceilometer service monitor, which is created above. The 2 service monitors need 2 distinct names.

Looking at the current contents and at the changes needed for TLS, I think it might be a good idea to create a new file in pkg/metricstorage for generating the ServiceMonitor structure for this.

controllers/ceilometer_controller.go
paramite: KSM is a stateless exporter without any storage and hence quite easily interchangable. I understand that it make sense to use statefulsets for more complex services, but in this case IMO Deployment is proper deployment method according to k8s documentation. Is there any other reason (which I currently don't see) other than consistency for which I should switch to StatefulSet?

vyzigold: No, consistency was the only reason.

jlarriba:
/hold

According to our latest conversations with team's steward, this feature is scheduled for FP1. With that in mind, and against what I would have loved (made possible by a great effort), I have to admit that delivering this starting pieces on GA without delivering the rest (for example, a ScrapeConfig for Prometheus to retrieve the metrics) will not be good for us in the long term.

paramite:
> (for example, a ScrapeConfig for Prometheus to retrieve the metrics) will not be good for us in the long term.

Scrape config creation is part of this PR via ServiceMonitor btw.

controllers/ceilometer_controller.go
```diff
@@ -452,7 +471,175 @@ func (r *CeilometerReconciler) reconcileNormal(ctx context.Context, instance *te
 		return ctrl.Result{}, err
 	}
 
-	Log.Info("Reconciled Service successfully")
+	Log.Info(fmt.Sprintf("Reconciled Service '%s' successfully", ceilometer.ServiceName))
+	return ctrl.Result{}, nil
+}
+
+func (r *CeilometerReconciler) reconcileKSM(
```
csibbitt, line 478: I'm wondering why all the KMS related stuff is being coupled with ceilometer? AFAICT it's a separate service with its own scrape endpoint and doesn't appear to have anything to do with ceilometer. What am I missing?

jlarriba: We dont want to have multiple APIs for every single component we deploy. The idea is for the user to deploy "blocks" of functionality using the operator. In this case, KMS is deployed along ceilometer because they both serve the same function: agents that gather metrics and expose prometheus metrics.

It is true that while this is evolving, the names of the controllers/APIs might get a little bit dated. In this case, we could consider this to be a "metricagents_controller" that takes care of all the control plane agents deployed. That could be the name, but I dont really see the need to couple the KMS development with the renaming of the API/controller. We could add the renaming as a low priority task in the Telemetry Refining epic.

csibbitt: Okay, so the name is just misleading, that makes sense given the history. I think this will eventually be a source of customer confusion, so I'd rate it a bit higher than "low priority", but I understand now, thanks.

elfiesmelfie:
> 
> https://review.rdoproject.org/zuul/buildset/16e2b0a889744e5a9834e0ed79647e53
> 
> ✔️ [openstack-k8s-operators-content-provider ](https://review.rdoproject.org/zuul/build/a83070139bd34677844200267a4d479b) SUCCESS in 1h 58m 19s ❌ [podified-multinode-edpm-deployment-crc ](https://review.rdoproject.org/zuul/build/6d1ceb29c6b749d19a82437bda3a4f07) **FAILURE** in 1h 03m 11s ❌ [telemetry-operator-multinode-autoscaling-tempest ](https://review.rdoproject.org/zuul/build/ec78796307c14fd09473183c64f956ed) **FAILURE** in 1h 33m 52s

telemetry-operator-multinode-autoscaling-tempest failed due to autoscaling.spec.aodh.memcachedInstance == '', which causes the deployment to time out waiting for a non-existant memcached instance to come up; This issue is resolved and the job should pass (hopefully) once you sync to main.

podified-multinode-edpm-deployment-crc failed for some reasons related to Telemetry. The control plane timed out deploying, and the telemetry CR has a "Waiting for ceilometer" status. I don't see any ceilometer CR, service, route or pod created though. The openstack-operators/pods/telemetry-controller-manager [1] might be a good place to look. However I don't know what that file typically looks like.

[1] https://logserver.rdoproject.org/37/337/36d131d752e096481d4bd51e3265e032d652f11f/github-check/podified-multinode-edpm-deployment-crc/6d1ceb2/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/telemetry-operator-controller-manager-5bd9bbd85f-82qwh/logs/manager.log

vyzigold:
I believe this will need to move to ScrapeConfigs too now. Fortunately it's quite simple and we have multiple examples. See: https://github.com/openstack-k8s-operators/telemetry-operator/pull/446 or https://github.com/openstack-k8s-operators/telemetry-operator/pull/442 or https://github.com/infrawatch/service-telemetry-operator/pull/607

jlarriba:
Its failing a legitimate kuttl test:

```
case.go:380: resource Service:telemetry-kuttl-tests/kube-state-metrics: .metadata.ownerReferences.name: value mismatch, expected: ceilometer != actual: telemetry-kuttl
    case.go:380: --- StatefulSet:telemetry-kuttl-tests/kube-state-metrics
        +++ StatefulSet:telemetry-kuttl-tests/kube-state-metrics
        @@ -2,20 +2,44 @@
         kind: StatefulSet
         metadata:
           labels:
        +    app.kubernetes.io/component: exporter
        +    app.kubernetes.io/name: kube-state-metrics
        +    app.kubernetes.io/version: v2.10.0
             service: kube-state-metrics
        +  managedFields: '[... elided field over 10 lines long ...]'
           name: kube-state-metrics
           namespace: telemetry-kuttl-tests
           ownerReferences:
        -  - kind: Ceilometer
        -    name: ceilometer
        +  - apiVersion: telemetry.openstack.org/v1beta1
        +    blockOwnerDeletion: true
        +    controller: true
        +    kind: Ceilometer
        +    name: telemetry-kuttl
        +    uid: 7dd2c187-64ce-4820-b6c2-9adb7a41c1bd
         spec:
        +  persistentVolumeClaimRetentionPolicy:
        +    whenDeleted: Retain
        +    whenScaled: Retain
        +  podManagementPolicy: OrderedReady
           replicas: 1
        +  revisionHistoryLimit: 10
           selector:
             matchLabels:
               app.kubernetes.io/name: kube-state-metrics
        +  serviceName: ""
        +  template: '[... elided field over 10 lines long ...]'
        +  updateStrategy:
        +    rollingUpdate:
        +      partition: 0
        +    type: RollingUpdate
         status:
           availableReplicas: 1
        +  collisionCount: 0
           currentReplicas: 1
        +  currentRevision: kube-state-metrics-c667d68b6
        +  observedGeneration: 1
           readyReplicas: 1
           replicas: 1
        +  updateRevision: kube-state-metrics-c667d68b6
        +  updatedReplicas: 1
```

paramite:
Hmm, ok I have no idea how https://github.com/paramite/telemetry-operator/blob/ksm/controllers/ceilometer_controller.go#L606 and https://github.com/paramite/telemetry-operator/blob/ksm/controllers/ceilometer_controller.go#L736 can have different result ... eg. one setting Ceilometer/ceilometer and the other one Ceilometer/telemetry-kuttl. BTW on my local env the test pass. Any hint what I am doing wrong?

jlarriba:
> Hmm, ok I have no idea how https://github.com/paramite/telemetry-operator/blob/ksm/controllers/ceilometer_controller.go#L606 and https://github.com/paramite/telemetry-operator/blob/ksm/controllers/ceilometer_controller.go#L736 can have different result ... eg. one setting Ceilometer/ceilometer and the other one Ceilometer/telemetry-kuttl. BTW on my local env the test pass. Any hint what I am doing wrong?

The don't have different result, both return "telemetry-kuttl". I believe that the problem is that in the assert yaml file, the StatefulSet for Ceilometer expects it to be "telemetry-kuttl" while the StatefulSet for kube-state-metrics expects "ceilometer", which is why it is complaning.

paramite:
Oh, ok I was confused with the ceilometer suite then. Prow is probably deploying resources different way than running with `make kuttl-test`. Maybe we might work on unifying, so that we avoid such confusion for other devs in future.

paramite:
Closing in favor of https://github.com/openstack-k8s-operators/telemetry-operator/pull/487.

PR #338, opened 2024-03-13
Pre commit update

paramite:
We are using old version of controller-gen.


vyzigold:
This would get us out of sync with the rest of the operators. I did a search across the openstack-k8s-operators org and all of the other operators are also using v0.11.1 except the test-operator. I believe this would cause issues when trying to develop more operators at once, because each of them would expect a different version of the controller-gen.

paramite:
True. I believe we should not use ancient software when developing new one :). This is just start.

paramite:
After discussing with folks: we cannot go to 0.14 as it needs [k8s.io](http://k8s.io/) 0.29 and we are on 0.28 due to golang version requirements.

PR #343, opened 2024-03-18
We cannot Own OBO resources in here because OBO might not exist in th…

jlarriba:
…e cluster when this is ran


jlarriba:
recheck

vyzigold:
> @jlarriba I think there may be more needed here than this. If I recall the telemetry-operator is error'ing out even if you don't create any resources. I think the issue is perhaps in main.go, where we register the same obo-prometheus-operator schema utilruntime.Must(monv1.AddToScheme(scheme))

That AddToScheme line shouldn't be the issue. It's there for like half a year. We had similar issues in the past, which were fixed by moving Owns() and Watches() of COO resources from the SetupWithManager() function to inside of the MetricStorage controller like this: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/controllers/metricstorage_controller.go#L218. But the line this PR removes unfortunately slipped the review and the CI.

The root cause is, that the SetupWithManager() functions are always called from main right when the operator starts and at that point, there might not be the COO CRDs and the whole operator fails. By moving the Owns() and Watches() calls inside the controller, we can look for the CRDs only when we have a MetricStorage CR in the cluster and react in the CRs status fields instead of erroring.

vyzigold:
This is still failing because of missing images:
```
  Normal   AddedInterface  19m                  multus             Add eth0 [10.130.0.69/23] from ovn-kubernetes
  Normal   Pulling         17m (x4 over 19m)    kubelet            Pulling image "quay.io/podified-antelope-centos9/openstack-heat-api@sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda"
  Warning  Failed          17m (x4 over 19m)    kubelet            Failed to pull image "quay.io/podified-antelope-centos9/openstack-heat-api@sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda": rpc error: code = Unknown desc = reading manifest sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda in quay.io/podified-antelope-centos9/openstack-heat-api: manifest unknown
  Warning  Failed          17m (x4 over 19m)    kubelet            Error: ErrImagePull
  Warning  Failed          17m (x6 over 19m)    kubelet            Error: ImagePullBackOff
  Normal   BackOff         4m7s (x63 over 19m)  kubelet            Back-off pulling image "quay.io/podified-antelope-centos9/openstack-heat-api@sha256:9a6bbbf99ae76ea9156566189e2cfa4140326036153a2a489a92a2d05688ffda"
```

csibbitt (approved):

vyzigold:
/retest
Another PR just merged, so the CI might finally work correctly.

vyzigold:
recheck
retry limit

jlarriba:
recheck

lets try once more and see if we get those resources...

vyzigold:
recheck
```
Traceback (most recent call last):
  File "/var/lib/tempest/external-plugins/.venv/lib64/python3.9/site-packages/telemetry_tempest_plugin/aodh/api/sql/test_alarming_api.py", line 75, in test_create_list_sort_limit_delete_alarm
    self.assertEqual(set(sevs),
  File "/var/lib/tempest/external-plugins/.venv/lib64/python3.9/site-packages/testtools/testcase.py", line 393, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/var/lib/tempest/external-plugins/.venv/lib64/python3.9/site-packages/testtools/testcase.py", line 480, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: {'critical', 'moderate'} != {'critical'}
```
in tempest TelemetryAlarmingAPIMysqlTest

PR #344, opened 2024-03-19
Start MetricStorage controller first

vyzigold:
This should make the telemetry-operator fail faster in case there is a bug in the code and it requires COO to run. It should make it easier for the CI to discover these bugs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -131,6 +131,17 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


vyzigold:
This didn't work. Even though I see the operator pod is erroring and restarting, it still managed to successfully deploy ceilometer in telemetry-operator-build-deploy

PR #349, opened 2024-03-21
Logging controller updates status instead of failing when cloNamespac…

jlarriba:
…e does not exist

Also, this raises the waiting time for triggering a new reconcile to 1 minute, 10 seconds does create a lot of spam in the logs and the user cannot really fix things in so little time.

Maybe the time could even be raised further to something like 5 minutes.


vyzigold:
Looks like this may be running into similar issues as my PR here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/348

PR #351, opened 2024-03-22
[zuul] Update the sg_core_url: ceilometer-internal -> ceilometer

elfiesmelfie:
The ceilometer-internal service is not always created

This is to help debug some downstream issues where ciloemter-internal is not available. This is to confirm the issue, and make sure there isn't some other issues blocking the tempest jobs from passing.
If this is the case, I need to add some way to set the sg_core_url so we can override it downstream without overriding the whole tempest conf

### DIFF
--- a/ci/vars-autoscaling-tempest.yml
+++ b/ci/vars-autoscaling-tempest.yml
@@ -5,6 +5,7 @@ cifmw_test_operator_tempest_container: openstack-tempest-all
 cifmw_test_operator_tempest_image_tag: 'current-podified'
 # This value is used to populate the `tempestconfRun` parameter of the Tempest CR: https://openstack-k8s-operators.github.io/test-operator/crds.html#tempest-custom-resource
 # https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/test_operator/defaults/main.yml
+# TODO: allow some way to select the sg_core URLwithout overriding the whole config
 cifmw_tempest_tempestconf_config:
   overrides: |
       validation.run_validation true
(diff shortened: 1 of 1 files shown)


vyzigold:
I don't understand why we should have "ceilometer" service downstream and not "ceilometer-internal". One thing why that may be happening is, that maybe we don't test with the latest T-O image downstream (I'm not sure how downstream testing works). The reason why I think that is, that the change which changed the service name didn't get as a dependency to openstack-operator yet. There are 3 PRs currentry which bump T-O in openstack-operator and they might be able to fix this issue: https://github.com/openstack-k8s-operators/openstack-operator/pull/710 https://github.com/openstack-k8s-operators/openstack-operator/pull/718 https://github.com/openstack-k8s-operators/openstack-operator/pull/716

One of them just needs to get through the CI, which isn't an easy thing to do lately.

elfiesmelfie:
recheck

elfiesmelfie:
recheck

elfiesmelfie:
> 
> https://review.rdoproject.org/zuul/buildset/3677185979124310bd54845f3bbf3ac3
> 
> ✔️ [openstack-k8s-operators-content-provider ](https://review.rdoproject.org/zuul/build/fe2a7b53206e43849910c277ad656e04) SUCCESS in 2h 04m 29s ❌ [podified-multinode-edpm-deployment-crc ](https://review.rdoproject.org/zuul/build/547f41b28ea142c7924ffc91da195f2a) **FAILURE** in 50m 30s ❌ [telemetry-operator-multinode-autoscaling-tempest ](https://review.rdoproject.org/zuul/build/09c69d9ad382436392be9ff5beb47f62) **FAILURE** in 50m 33s

Unknown failure, but also unrelated to telemetry

elfiesmelfie:
Okay, so the ceilometer-internal service is created, but no ceilometer service. I'm going to close this. Since ceilometer-internal is correct
I'll disable the test downstream for now and investigate the differences when I have the job enabled.

PR #353, opened 2024-03-25
[zuul] Add a job to check operator errors

vyzigold:
This checks for errors in the operator logs.

For testing purposes I introduced an error in the metricstorage controller, so the job should fail right now.

### DIFF
--- a/ci/vars-default-telemetry.yml
+++ b/ci/vars-default-telemetry.yml
@@ -1,7 +1,7 @@
 ---
-cifmw_edpm_prepare_timeout: 60
-pre_deploy_create_coo_subscription:
-    source: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/create-coo-subscription-playbook.yaml"
+post_ctlplane_deploy:
+  - name: Check default telemetry
+    source: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/check-default-telemetry.yml"
     type: playbook
 cifmw_edpm_prepare_kustomizations:
   - apiVersion: kustomize.config.k8s.io/v1beta1
@@ -18,12 +18,5 @@ cifmw_edpm_prepare_kustomizations:
             enabled: true
           telemetry:
             enabled: true
-            template:
-              metricStorage:
-                enabled: true
-                monitoringStack:
-                  alertingEnabled: false
-              autoscaling:
-                enabled: true
       target:
         kind: OpenStackControlPlane
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -18,6 +18,17 @@
         - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-autoscaling.yml"
         - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-autoscaling-tempest.yml"
 
+- job:
+    name: telemetry-operator-multinode-default-telemetry
+    parent: podified-multinode-edpm-deployment-crc
+    dependencies: ["openstack-k8s-operators-content-provider"]
+    description: |
+      Deploy a default OpenStack with telemetry enabled, but without COO installed. Check that the telemetry-operator logs don't have any errors.
+    vars:
+      cifmw_extras:
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir  }}/scenarios/centos-9/multinode-ci.yml"
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-default-telemetry.yml"
+
 - project-template:
     name: rdo-telemetry-tempest-plugin-jobs
     openstack-experimental:
@@ -54,3 +65,4 @@
     github-check:
       jobs:
         - telemetry-operator-multinode-autoscaling-tempest
+        - telemetry-operator-multinode-default-telemetry
--- a/ci/check-default-telemetry.yml
+++ b/ci/check-default-telemetry.yml
@@ -49,11 +49,8 @@
             msg: "Errors were found in the telemetry-operator logs"
       when: error_list | length != 0
 
-    - name: "Get telemetry-operator container restart counts"
-      ansible.builtin.command:
-        cmd:
-          oc get pods -n openstack-operators -l "openstack.org/operator-name=telemetry" --output=jsonpath --template='{.items[*].status.containerStatuses[*].restartCount}'
-      register: restart_counts
+    - name: "Get telemetry-operator restart counts"
+      ansible.builtin.import_tasks: "get-operator-restart-counts.yml"
 
     - name: "Fail if telemetry-operator was restarted"
       block:
@@ -70,7 +67,7 @@
         - name: "Fail because telemetry-operator restarted"
           ansible.builtin.fail:
             msg: "Telemetry-operator restarted"
-      when: restart_counts.stdout != "0 0"
+      when: restart_counts != [0, 0]
 
     # Enabling Autoscaling and MetricStorage without COO installed. Expected to see errors.
     - name: "Enable MetricStorage"
@@ -88,11 +85,8 @@
       ansible.builtin.wait_for:
         timeout: 120
 
-    - name: "Get telemetry-operator container restart counts"
-      ansible.builtin.command:
-        cmd:
-          oc get pods -n openstack-operators -l "openstack.org/operator-name=telemetry" --output=jsonpath --template='{.items[*].status.containerStatuses[*].restartCount}'
-      register: restart_counts
+    - name: "Get telemetry-operator restart counts"
+      ansible.builtin.import_tasks: "get-operator-restart-counts.yml"
 
     - name: "Fail if telemetry-operator was restarted"
       block:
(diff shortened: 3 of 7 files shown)


zuul.d/projects.yaml
```diff
@@ -29,6 +29,17 @@
       # This value will be used when the telemetry-tempest-plugin is released, and we no longer need to install from the master branch
       # cifmw_test_operator_tempest_external_plugin: "{{ [] if change_item | length < 1 else [ { 'repository': 'https://' + external_plugin + '.git', 'changeRepository': 'https://review' + {{ external_plugin }}, 'changeRefspec': [ 'refs/changes', change_item[0].change[-2:], change_item[0].change, change_item[0].patchset ] | join('/') } ] }}"
 
+- job:
+    name: telemetry-operator-multinode-check-errors
```
elfiesmelfie, line 33: The job name should be updated and the description added.

This job tests the expected default deployment of telemetry-operator i.e. telemetry enabled, metricsstorage enabled, no coo.
We want to make sure that the failure is graceful, rather than throwing an exception.

The name should be something like `telemetry-operator-multinode-default-telemetry` or something like that.
The description should include that we are testing the default behaviour, and making sure that metricsStorage fails gracefully and waits for COO if it is not installed.

ci/vars-check-errors.yml
```diff
@@ -0,0 +1,5 @@
+---
+cifmw_run_test_role: test_operator
```
elfiesmelfie, line 2: I don't think this is needed.

You're not using the test_operator is used to run either tempest or tobiko (https://github.com/openstack-k8s-operators/ci-framework/blob/8d9bbe7c7cbba3eb900263d03f9d59e3203b1951/docs/source/usage/01_usage.md?plain=1#L44)
The default is to run tempest via test_operator. 
In this case, the tempest tests would not be useful, since the default config would run all the tests from core tempest, rather than from tempest-plugins.
https://github.com/openstack-k8s-operators/ci-framework/blob/8d9bbe7c7cbba3eb900263d03f9d59e3203b1951/roles/test_operator/README.md
The cifmw_test_operator_tempest_container defaults to openstack-tempest, which installs the tempest package, but not the tempest plugins from the `openstack-tempest-all` package, which includes telemetry-tempest-plugin.

Even if you configure tempest to run, I think there's going to be only one or two that work, since only  ceilometer and sg_core will be available.

ci/check-errors.yml
```diff
@@ -0,0 +1,32 @@
+---
```
elfiesmelfie, line 1: Based on the other comments, this should be renamed, and also check for expected behaviour. 

e.g. Does the MetricStorage report the right status?

elfiesmelfie (changes requested):

ci/check-errors.yml
vyzigold: MetricStorage shouldn't get deployed in this scenario, so there is nothing to check (the operator can still fail though). But I'll check the status of telemetry. Thanks.

jlarriba: I believe that MetricStorage should get deployed in this scenario, because that way we will be not only detecting the error you are trying here (that we added an Owns over a COO API), but also test that MetricStorage fails gracefully without bringing the whole operator down. This test case should test those two things at once, instead of only one of them.

vyzigold: Alright, I'll try to add MetricStorage to the deployment.

zuul.d/projects.yaml
vyzigold: I'm not sure what you mean by the "graceful failure". I guess I shouldn't use the `ansible.builtin.fail`? What are the alternatives then?

ci/check-errors.yml
vyzigold: While I was at it, I also enabled autoscaling, so we can deploy the whole telemetry. I am a little unsure about enabling MetricStorage though. In my opinion, if COO isn't installed, then MetricStorage shouldn't get to a Ready state. That means Telemetry won't get to Ready state and I think because of that the control-plane won't get to the ready state and the deployment won't be successful. But lets see what the CI thinks about this first.

vyzigold:
cert-manager issues. I've seen that on another PR as well.

ci/vars-default-telemetry.yml
```diff
@@ -0,0 +1,28 @@
...
+          heat:
+            enabled: true
+          telemetry:
+            enabled: true
+            template:
+              metricStorage:
+                enabled: true
+                monitoringStack:
+                  alertingEnabled: false
+              autoscaling:
```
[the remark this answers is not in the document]
vyzigold: We've discussed this with Juan in slack today, you can take a look for the details there. I think we settled on the idea to enable the whole telemetry without COO and check the logs and CR status fields in detail to verify everything is as expected - the operator is running, ceilometer is ready, metric storage isn't ready, but reports the correct statuses and is trying to look for COO.

Adding autoscaling to the deployment probably doesn't add much at the moment. We'd basically just check, that it's ready. But I don't think it hurts us in any way.

ci/default-telemetry.yml
```diff
@@ -0,0 +1,46 @@
+---
```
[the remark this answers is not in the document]
vyzigold: I wasn't completely sure about the naming this morning, but I ended up following what I thought you're suggesting :D . But I guess if we have a "vars-default-telemetry.yml" and "check-default-telemetry.yml", we could then be adding other "something-default-telemetry.yml" when we see the need. That makes sense.

vyzigold:
recheck

vyzigold:
Just as I worried. The test failed. CI framework didn't execute our tests because control plane wasn't ready because telemetry wasn't ready because metric-storage wasn't ready which is caused by missing COO. I'll temporarily disable metric-storage to see if the rest works OK and if CI framework executes our tests.

vyzigold:
recheck
It doesn't seem like it triggered for some reason.

elfiesmelfie:
> Just as I worried. The test failed. CI framework didn't execute our tests because control plane wasn't ready because telemetry wasn't ready because metric-storage wasn't ready which is caused by missing COO. I'll temporarily disable metric-storage to see if the rest works OK and if CI framework executes our tests.

You can also add the check playbook to [Zuul's post-run config](https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.post-run) while you're debugging and fixing the issue. 
You can leave the playbook as a hook, but also add it as a post-run playbook temporarily, which will always be executed in a job, and help to debug and verify any fixes you put in.

vyzigold:
It's waiting for heat to get ready. We need to wait until a bug in heat is fixed.

vyzigold:
recheck

elfiesmelfie:
> 
> https://review.rdoproject.org/zuul/buildset/178dc9387a44478bbfe6ebbd9068dfa8
> 
> ✔️ [openstack-k8s-operators-content-provider ](https://review.rdoproject.org/zuul/build/3c0dd576baa34b829801ac98029ff485) SUCCESS in 1h 56m 54s ✔️ [podified-multinode-edpm-deployment-crc ](https://review.rdoproject.org/zuul/build/07b78ad49a454a1faddcdc5b5bc4049b) SUCCESS in 1h 14m 54s ❌ [telemetry-operator-multinode-autoscaling-tempest ](https://review.rdoproject.org/zuul/build/c928a55166974bbb8845e8c99b008472) **FAILURE** in 1h 33m 48s ✔️ [telemetry-operator-multinode-default-telemetry ](https://review.rdoproject.org/zuul/build/33b88be10b9c4b0d8549b2b765e9571d) SUCCESS in 1h 11m 04s

The failure is due to the (memcached instance not being provisioned)[https://logserver.rdoproject.org/53/353/f0772c3da4fe5f88b32f96ee5c21baf2309c4806/github-check/telemetry-operator-multinode-autoscaling-tempest/c928a55/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/autoscalings.telemetry.openstack.org/autoscaling.yaml].
The spec.aodh.memcachedInstance = "" i.e. it is not being set.

Previously (see [here](https://logserver.rdoproject.org/53/353/f0772c3da4fe5f88b32f96ee5c21baf2309c4806/github-check/telemetry-operator-multinode-autoscaling-tempest/ca9b0de/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/autoscalings.telemetry.openstack.org/autoscaling.yaml)), the instance was set to `memcached`. There's also a bunch of other differences in the config, including the spec.aodh.override being set previously, but not now.

vyzigold:
Hmm, that's strange. telemetry-operator-multinode-autoscaling-tempest isn't really something I'd normally care on this PR at the moment. But memcached not being set almost seems like we may have some kind of bug somewhere :thinking:

elfiesmelfie:
> Hmm, that's strange. telemetry-operator-multinode-autoscaling-tempest isn't really something I'd normally care on this PR at the moment. But memcached not being set almost seems like we may have some kind of bug somewhere 🤔

Definitely a bug somewhere, I saw this issue earlier in the week, and Marihan is also seeing it here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/353

elfiesmelfie:
> 
> https://review.rdoproject.org/zuul/buildset/ccfeaddc341b429eae31f29d445a0f3e
> 
> ✔️ [openstack-k8s-operators-content-provider ](https://review.rdoproject.org/zuul/build/110df85772364107a5e19fff6f27d3c5) SUCCESS in 1h 55m 26s ✔️ [podified-multinode-edpm-deployment-crc ](https://review.rdoproject.org/zuul/build/026eaa454797418bae35fd38230ec668) SUCCESS in 1h 11m 43s ❌ [telemetry-operator-multinode-autoscaling-tempest ](https://review.rdoproject.org/zuul/build/be38d12e9fd84df3ad7e494ec3d9ac0d) **FAILURE** in 1h 33m 24s ✔️ [telemetry-operator-multinode-default-telemetry ](https://review.rdoproject.org/zuul/build/03652bdab981485f942d77129b60b5ee) SUCCESS in 1h 09m 27s

The failure was due to spec.aodh.memcachedInstance == ''. This issue has been resolved.

elfiesmelfie:
recheck

vyzigold:
I rewrote most of the ansible, so I squashed it and rebased on top of main.

elfiesmelfie:
recheck

vyzigold:
recheck
POST_FAILURE

vyzigold:
I forgot that I deliberately included a bug in the PR. I can confirm, that the job successfully detected the operator restarting.

vyzigold:
The autoscaling tempest job is failing because of `No resources found in openshift-operators namespace.` This isn't related to this PR.

jlarriba:
/lgtm

Awesome work done here!

PR #356, opened 2024-03-28
Update template for telemetry tempest plugin

elfiesmelfie:
Dependency listed is to verify the fix works, sine the jobs will be the same otherwise

Depends-On: https://review.opendev.org/c/openstack/telemetry-tempest-plugin/+/912723


elfiesmelfie:
recheck

jlarriba (approved):

PR #360, opened 2024-04-03
Improve defaulting

vyzigold:
This makes enables us to deploy a working telemetry with default values with minimal CR:

```
telemetry:
  enabled: true
  autoscaling:
    enabled: true
  ceilometer:
    enabled: true
  metricStorage:
    enabled: true
```
    
Depends-On: https://github.com/openstack-k8s-operators/heat-operator/pull/344

### DIFF
--- a/api/v1beta1/metricstorage_webhook.go
+++ b/api/v1beta1/metricstorage_webhook.go
@@ -51,6 +51,9 @@ func (spec *MetricStorageSpec) Default() {
 		// and we don't have a way of distinguishing if the value was left
 		// empty or set to false by the user in the spec.MonitoringStack.Default()
 		spec.MonitoringStack.AlertingEnabled = true
+
+		// NOTE: If we want to enable dashboards in the future by default, set
+		//       it here like Alerting above
 	}
 	if spec.MonitoringStack != nil {
 		spec.MonitoringStack.Default()
@@ -61,6 +64,9 @@ func (spec *MetricStorageSpec) Default() {
 func (ms *MonitoringStack) Default() {
 	if ms.ScrapeInterval == "" {
 		ms.ScrapeInterval = "30s"
+	}
+	if ms.DataplaneNetwork == "" {
+		ms.DataplaneNetwork = "ctlplane"
 	}
 	ms.Storage.Default()
 }
--- a/api/bases/telemetry.openstack.org_autoscalings.yaml
+++ b/api/bases/telemetry.openstack.org_autoscalings.yaml
@@ -36,6 +36,18 @@ spec:
             description: AutoscalingSpec defines the desired state of Autoscaling
             properties:
               aodh:
+                default:
+                  apiImage: ""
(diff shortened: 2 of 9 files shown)


jlarriba:
recheck

vyzigold:
recheck

api/v1beta1/metricstorage_webhook.go
```diff
@@ -51,6 +51,9 @@ func (spec *MetricStorageSpec) Default() {
 		// and we don't have a way of distinguishing if the value was left
 		// empty or set to false by the user in the spec.MonitoringStack.Default()
 		spec.MonitoringStack.AlertingEnabled = true
+
+		// NOTE: If we want to enable dashboards in the future by default, set
+		//       it here like Alerting above
```
csibbitt, line 56: :+1:

csibbitt (approved):
What is missing for us to actually enable ceilometer, metricStorage, and autoscaling by default?

jlarriba:
recheck

jlarriba:
recheck

vyzigold:
> What is missing for us to actually enable ceilometer, metricStorage, and autoscaling by default?

I don't think there are plans to have autoscaling enabled by default. With metricStorage the issue is the COO dependency. We want to tell the users to always enable it, it should be enabled in all the VAs, but if we enable it here, the CI in all of the other operators will fail, because metricStorage will get deployed, but it'll never get to a ready state without COO.

vyzigold:
recheck

jlarriba:
recheck

csibbitt:
> > What is missing for us to actually enable ceilometer, metricStorage, and autoscaling by default?
> 
> I don't think there are plans to have autoscaling enabled by default. With metricStorage the issue is the COO dependency. We want to tell the users to always enable it, it should be enabled in all the VAs, but if we enable it here, the CI in all of the other operators will fail, because metricStorage will get deployed, but it'll never get to a ready state without COO.

Thanks for this, Jaromir. I hadn't fully understood where the "real" defaults would be set (Verified Architecture manifests), vs the defaults that need to be in use for CI. This helps!

Ultimately, I hope that we find a supportable way of installing COO (and other operator bundles we depend on) the way we do with RPMs.

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
Hmm, looks like the failures are actual real failures because of the PR. I'll need to look into it.

vyzigold:
This tries to default memcached, which is getting removed here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/365 Once 365 gets merged, I'll get back to this, rebase it and remove the memcached defaulting.

elfiesmelfie:
> This tries to default memcached, which is getting removed here: #365 Once 365 gets merged, I'll get back to this, rebase it and remove the memcached defaulting.

This has been resolved and other PRs are passing now

PR #365, opened 2024-04-16
[OSPRH-6312] Remove memcached from autoscaling

vyzigold:
During investigatin of issues with [OSPRH-6312](https://issues.redhat.com//browse/OSPRH-6312) we discovered, that memcached in fact isn't required for aodh. Because it was a required parameter, it actually caused additional unnecessary issues. This PR removes memcached from the autoscaling api, from all example CRs and from kuttl tests.

rh-jira: https://issues.redhat.com/browse/OSPRH-6312

### DIFF
--- a/api/bases/telemetry.openstack.org_autoscalings.yaml
+++ b/api/bases/telemetry.openstack.org_autoscalings.yaml
@@ -70,10 +70,6 @@ spec:
                     type: string
                   listenerImage:
                     type: string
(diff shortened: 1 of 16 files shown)


paramite:
Did not test this, but code looks legit.

jlarriba:
let's see how it runs through CI

jlarriba:
recheck

PR #366, opened 2024-04-17
Manage ObservabilityUIPlugin for Dashboards plugin

csibbitt:
* We need to control a new CR that installs the plugin
* Using Unstructured{} to avoid golang 1.21 and core k8s dep updates in newest OBO

### DIFF
--- a/controllers/metricstorage_controller.go
+++ b/controllers/metricstorage_controller.go
@@ -492,6 +492,12 @@ func (r *MetricStorageReconciler) reconcileNormal(
 		instance.Status.Conditions.MarkTrue(telemetryv1.DashboardDefinitionReadyCondition, telemetryv1.DashboardsNotEnabledMessage)
 		instance.Status.Conditions.MarkTrue(telemetryv1.DashboardPluginReadyCondition, telemetryv1.DashboardsNotEnabledMessage)
 	} else {
+
+		const (
+			dashboardArtifactsNamespace = "openshift-config-managed"
+			dashboardPluginNamespace    = "openshift-operators"
+		)
+
 		// Deploy dashboard UI plugin from OBO
 		// TODO: Use the following instead of Unstructured{} after COO 0.2.0
 		// =====
@@ -514,12 +520,11 @@ func (r *MetricStorageReconciler) reconcileNormal(
 			Kind:    "ObservabilityUIPlugin",
 		})
 		uiPluginObj.SetName("ui-dashboards")
-		uiPluginObj.SetNamespace(instance.Namespace)
+		uiPluginObj.SetNamespace(dashboardPluginNamespace)
 		// =====
 		op, err = controllerutil.CreateOrPatch(ctx, r.Client, uiPluginObj, func() error {
 			// uiPluginObj.Spec.Type = "Dashboards" // After COO 0.2.0
-			err = controllerutil.SetControllerReference(instance, uiPluginObj, r.Scheme)
-			return err
+			return nil
 		})
 		if err != nil {
 			Log.Error(err, fmt.Sprintf("Failed to update Dashboard Plugin definition %s - operation: %s", uiPluginObj.GetName(), string(op)))
@@ -572,13 +577,17 @@ func (r *MetricStorageReconciler) reconcileNormal(
 		datasourceCM := &corev1.ConfigMap{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      datasourceName,
-				Namespace: instance.Namespace,
+				Namespace: "console-dashboards", // Should be "dashboardArtifactsNamespace" once https://github.com/openshift/console-dashboards-plugin/pull/35 is in the build
 			},
 		}
 		dataSourceSuccess := false
 		op, err = controllerutil.CreateOrPatch(ctx, r.Client, datasourceCM, func() error {
 			datasourceCM.ObjectMeta.Labels = map[string]string{
 				"console.openshift.io/dashboard-datasource": "true",
+			}
+			scheme := "http"
+			if instance.Spec.PrometheusTLS.Enabled() {
+				scheme = "https"
 			}
 			datasourceCM.Data = map[string]string{
 				// WARNING: The lines below MUST be indented with spaces instead of tabs
@@ -590,7 +599,7 @@ func (r *MetricStorageReconciler) reconcileNormal(
                         plugin:
                             kind: "PrometheusDatasource"
                             spec:
-                                direct_url: "http://prometheus-operated.` + instance.Namespace + ".svc.cluster.local:9090\"",
+                                direct_url: "` + scheme + `://metric-storage-prometheus.` + instance.Namespace + ".svc.cluster.local:9090\"",
 			}
 			return nil
 		})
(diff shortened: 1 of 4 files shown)


controllers/metricstorage_controller.go
```diff
@@ -529,7 +572,7 @@ func (r *MetricStorageReconciler) reconcileNormal(
 		datasourceCM := &corev1.ConfigMap{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      datasourceName,
-				Namespace: "console-dashboards",
+				Namespace: instance.Namespace,
```
csibbitt, line 575: Previously we were installing the plugin via helm with `-n console-dashboards --create-namespace`. Now we request the plugin in whatever namespace our operator runs in (openstack-operators), and the datasource object lives with it there.

There is a possibility that this may need to change, as there was discussion about how to handle multiple ObservabilityUIPlugins (OUIPs), and that perhaps we'd need to settle on a single canonical namespace and somehow co-ordinate shared ownership of the OUIPs within that namespace. I'm watching https://issues.redhat.com/browse/OU-402 for updates.

@jgbernalp does that sound about right, or is there something I should do now to make sure this doesn't break later?

csibbitt:
I tested this like so:
* install observability-operator.v0.2.0-dev
* run T-O from local source
* deploy config/samples/telemetry_v1beta1_telemetry.yaml
* edit telemetry object to enable metricStorage and dashboards
* observe that the dashboards appear in the console and panels are populating

controllers/metricstorage_controller.go
csibbitt: Ooops, so actually more testing reveals that this is wrong and the datasource doesn't work. I think it's because that namespace name is hard coded in the plugin?

["The plugin will search for datasources as ConfigMaps in the `console-dashboards` namespace (plugin docs)"](https://github.com/openshift/console-dashboards-plugin/blob/main/docs/add-datasource.md?plain=1#L3)

OBO doesn't appear to create this namespace, though, so I guess it's on us to create it if it doesn't already exist, and put both the OUIP (if it doesn't already exist) and the datasource in there?

I'll figure this out tomorrow.

csibbitt:
## Tests failed

Okay, upon closer inspection it seems like the datasource is actually not working and the panels are populating with data from the default (openshift-monitoring) datasource :frowning_face: 

A bit more detail in this comment https://github.com/openstack-k8s-operators/telemetry-operator/pull/366#discussion_r1568077619

controllers/metricstorage_controller.go
jlarriba: Normally, openshift operators are fixed to a certain namespace. cluster-logging can only work on "openshift-logging" namespace and will not work anywhere else, so this could perfectly be the case.

jan--f:
@csibbitt Just as a heads up: The final release of this will likely have a different GVK name: https://github.com/rhobs/observability-operator/pull/434/files#diff-87f3ff0ffd8a86bfc90125fa3543caff45c234d6d2d54e2da371c560b65c0e1dR129

controllers/metricstorage_controller.go
csibbitt: In this case my concern is not that we need to _use_ a particular namespace, but rather that we need to _create it_. I'll add the code because I want this to be working asap, but it seems wrong. If this namespace is hard-coded into the plugin, I'd expect the plugin (or OBO) to manage it instead of our code.

csibbitt:
> @csibbitt Just as a heads up: The final release of this will likely have a different GVK name: https://github.com/rhobs/observability-operator/pull/434/files#diff-87f3ff0ffd8a86bfc90125fa3543caff45c234d6d2d54e2da371c560b65c0e1dR129

Thanks!

controllers/metricstorage_controller.go
jgbernalp: Hi @csibbitt the namespace can be configured in the plugin, but this option is not yet available using COO. I suggest to use the same namespace as the dashboards `openshift-config-managed` to store the datasources as the plugin can read from there and this will become the default.

csibbitt: Just to confirm; if I alter my code to make it use `openshift-config-managed` right now it doesn't work, but if I copy the datasource CM to `console-dashboards`, it does. I'm assuming the next dev bundle build I get my hands on will change the default.

csibbitt:
## Testing Progress
This is working now provided I do these two things:

1. Manually create "console-dashboards" namespace. This is flagged in the code as something that will change imminently
2. Turn off TLS in openstack (spec.tls.podLevel.enabled = false in the OpenStackControlPlane object)

controllers/metricstorage_controller.go
csibbitt: I experimented with different values for this today:
- "openshift-operators" - Maybe the best choice to standardize on? Works
- "openstack" (instance.Namespace) - Will lead to duplicate Objects from multiple projects, probably a bad choice. Works
- "console-dashboards" -  nothing creates it, and it's being abandoned in favor of "openshift-config-managed" as the default for other objects. Works (if you create it first)
- "openshift-config-managed" - probably not canonically correct but I tried it. Doesn't work - pod is stuck in "CreateContainerConfigError" with "Error: container has runAsNonRoot and image will run as root

CC: @jgbernalp @lihongyan1 We talked about this on the 4th (OU-402) - does this seem like the right choice to you?

jgbernalp: @csibbitt we decided to use the `openshift-config-managed` to have the dashboards and datasources in the same place. COO will configure the necessary permissions for the plugin to watch this namespace regardless of the namespace in which the plugin lives. The plugin will live in the same namespace as the COO operator. The new GKV UIPlugin is a cluster scoped resource and we will add validations to only have one instance per type.

csibbitt:
## Testing looks good

This latest code is tested against quay.io/gbernal/observability-operator-bundle:0.2.0-dev-1

* I can turn the global TLS on/off in openstack and see the datasource definition respond (full/empty CAcerts, and correct URL scheme)
* Dashboards are confirmed working (*with* the custom datasource, not the default fallback) in both modes and after switching between them multiple times

pkg/metricstorage/dashboard_datasource.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+func DashboardDatasourceData(ctx context.Context, c client.Client, instance *telemetryv1.MetricStorage, datasourceName string, namespace string) (map[string]string, error) {
+
+	scheme := "http"
+	certText := ""
+	if instance.Spec.PrometheusTLS.Enabled() {
+		scheme = "https"
+		namespacedName := types.NamespacedName{
+			Name:      instance.Spec.PrometheusTLS.CaBundleSecretName,
```
csibbitt, line 35: This works, but it doesn't seem strictly correct to me. I think technically this is the cert bundle that prometheus will use to validate its outbound connections (for example to scrape endpoints). What we actually want is the CA bundle for validating the cert that prometheus uses to serve its api. Right now they are one and the same, but they might not always be. I did a quick check to see if we allow users to supply their own serving certs to prometheus in a way that would break this and I think it's okay, but I'll admit it wasn't a thorough check yet.

CC: @vyzigold

vyzigold: In the instance.Spec.PrometheusTLS.secretName is "ca.crt" I wonder if that is the CA certificate to use.

vyzigold: Yes, it works for me when curling. I'm not sure if you'd be able to configure it here though.

vyzigold: Ah, reading a few lines further I think you should be able to configure it. It would be just:
L35: `Name:      instance.Spec.PrometheusTLS.secretName,`
L43: `certText = string(caSecret.Data["ca.crt"])`

csibbitt: Ah, thanks! It didn't occur to me that it would be included in that secret. That's definitely the better place to get it from, and I just confirmed they are identical. Adjustment incoming.

PR #323, opened 2024-02-29
 [zuul] Update the external_plugin var

mgirgisf:

PR #369, opened 2024-04-18
[zuul] Use https in sg_core_service_url

elfiesmelfie:
This fixes the tempest failure:

    Your browser sent a request that this server could not understand.<br />
    Reason: You're speaking plain HTTP to an SSL-enabled server port.<br />
    Instead use the HTTPS scheme to access this URL, please.<br />

The sg_core_service_url is passed straight through to the test and used as the URL base for HTTP requests. The https:// needs to be prefixed or else the connection will default to http, which is an issue when we have TLS everywhere.

### DIFF
--- a/ci/vars-autoscaling-tempest.yml
+++ b/ci/vars-autoscaling-tempest.yml
@@ -12,7 +12,7 @@ cifmw_tempest_tempestconf_config:
       service_available.ceilometer true
       service_available.sg_core true
       service_available.aodh true
-      telemetry.sg_core_service_url "ceilometer-internal.openstack.svc.cluster.local:3000"
+      telemetry.sg_core_service_url "https://ceilometer-internal.openstack.svc.cluster.local:3000"
 cifmw_test_operator_tempest_include_list: |
   telemetry_tempest_plugin.scenario
   telemetry_tempest_plugin.aodh


elfiesmelfie:
The issues came up here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/323#issuecomment-2061146196

ci/vars-autoscaling-tempest.yml
```diff
@@ -12,7 +12,7 @@ cifmw_tempest_tempestconf_config:
       service_available.ceilometer true
       service_available.sg_core true
       service_available.aodh true
-      telemetry.sg_core_service_url "ceilometer-internal.openstack.svc.cluster.local:3000"
+      telemetry.sg_core_service_url "https://ceilometer-internal.openstack.svc.cluster.local:3000"
```
csibbitt, line 15: I think http vs. https might need to be selected depending on whether TLS is enabled in the control plane (spec.tls.podLevel.enabled in the OpenStackControlPlane object). Not sure if this matters for this testing or if we can safely assume we're running with TLS enabled.

mgirgisf: we can update it for example to : sg_core_service_url: "{{ protocol }}://ceilometer-internal.openstack.svc.cluster.local:3000" , and we set the protocol in the job to https , but as emma mentioned its not mandatory now.

mgirgisf (approved):

PR #372, opened 2024-04-22
[OSPRH-6400] Add Proxy image to telemetry and default_images

vyzigold:
openstack-operator isn't able to set the image correctly and it isn't able to put it inside the openstackversions CR because it's missing in these files.

rh-jira: https://issues.redhat.com/browse/OSPRH-6400

### DIFF
--- a/config/default/manager_default_images.yaml
+++ b/config/default/manager_default_images.yaml
@@ -21,6 +21,8 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-ceilometer-ipmi:current-podified
         - name: RELATED_IMAGE_CEILOMETER_SGCORE_IMAGE_URL_DEFAULT
           value: quay.io/infrawatch/sg-core:v5.2.0-nextgen
+        - name: RELATED_IMAGE_CEILOMETER_PROXY_IMAGE_URL_DEFAULT
+          value: quay.io/podified-antelope-centos9/openstack-aodh-api:current-podified
         - name: RELATED_IMAGE_AODH_API_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-aodh-api:current-podified
         - name: RELATED_IMAGE_AODH_EVALUATOR_IMAGE_URL_DEFAULT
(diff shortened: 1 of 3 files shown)


config/default/manager_default_images.yaml
```diff
@@ -21,6 +21,8 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-ceilometer-ipmi:current-podified
         - name: RELATED_IMAGE_CEILOMETER_SGCORE_IMAGE_URL_DEFAULT
           value: quay.io/infrawatch/sg-core:v5.2.0-nextgen
+        - name: RELATED_IMAGE_CEILOMETER_PROXY_IMAGE_URL_DEFAULT
+          value: quay.io/podified-antelope-centos9/openstack-aodh-api:current-podified
```
csibbitt, line 25: I thought this must be a copy/paste error because it's the same image as L27 and I was expecting some generic https proxy image. I double checked and see that this is indeed the correct image (at least it's the one running in my environment), but it leaves me with a couple questions:

1. What's up with the name? "aodh-api" doesn't seem like the correct description for what appears to be a minimal https proxy setup.
2. If it's truly supposed to be the  same container as the one on the next line, why would we not just use that same variable in both places? I mean, the simple answer is that "RELATED_IMAGE_AODH_API_IMAGE_URL_DEFAULT" would also really out of place in the ceilometer deployment.

Is the answer to both questions that we have simply re-used the aodh-api container for an entirely different purpose by mounting a different config-data volume and repurposing the httpd to do TLS proxying for ceilometer?

vyzigold: You are correct. The intention is to reuse the image to get httpd, so that we can terminate the TLS connection. I don't think we want to reuse the RELATED_IMAGE_AODH_API_IMAGE_URL_DEFAULT for this, because these variables are used to generate the "openstackversions" CR, which is where the user could change these images (so we need a separate env variable). But let me ask on slack if we have a "general purpose" httpd image we could use or if reuse of aodh-api is fine.

vyzigold: I was pointed to: https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/293e48aceb9b704fe93bc438849393ab4cc49f7d/config/default/manager_default_images.yaml#L18-L19 . Which is a plain httpd image. I could try using that instead if you think it'll be better.

csibbitt: I don't think it makes much difference and it's probably better to be on an image that we control than borrowing one from the baremetal operator.

csibbitt (approved):

PR #373, opened 2024-04-22
Update telemetry-approvers

vyzigold:
- Emma is active in contributing to the operator recently, but I noticed she is missing from the telemetry approvers list.

- Matthias requested to be removed, because he doesn't contribute.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -3,8 +3,8 @@
 aliases:
   telemetry-approvers:
   - csibbitt
+  - elfiesmelfie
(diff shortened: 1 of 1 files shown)


mrunge (approved):
yes, definitely!

mrunge:
Btw, I see that I am listed there as well, but I do not contribute at all. On the same note Emma should be included, I think, I should be excluded.

vyzigold:
@mrunge I removed you from the list and updated the PR name and description accordingly.

PR #374, opened 2024-04-24
Ensure golangci-lint runs on all files

gibizer:
* removed unused func params and import
* ignore revive dot-imports rule on ginkgo and gomega as dot import
there is the recommended practice
* various small style fixes found by the linter

### DIFF
--- a/pkg/metricstorage/dashboard_prometheus_rule.go
+++ b/pkg/metricstorage/dashboard_prometheus_rule.go
@@ -26,7 +26,7 @@ import (
 func DashboardPrometheusRule(
 	instance *telemetryv1.MetricStorage,
 	labels map[string]string,
-	selector map[string]string,
+	_ map[string]string,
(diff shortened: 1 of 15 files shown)


pkg/metricstorage/dashboard_prometheus_rule.go
```diff
@@ -26,7 +26,7 @@ import (
 func DashboardPrometheusRule(
 	instance *telemetryv1.MetricStorage,
 	labels map[string]string,
-	selector map[string]string,
+	_ map[string]string,
```
gibizer, line 29: I did not removed this unused parameter as I could not decide if this is a bug that such param is not passed forward. Could somebody with more context check it please?

jlarriba: @csibbitt ^

jlarriba:
/approve /lgtm

We will be looking at the issue in a separate bug.

PR #390, opened 2024-05-09
[zuul] Disable autoscaling tempest scenario

vyzigold:
The scenario needs caracal images to succeed. Disable it until the CI starts using newer images.


zuul.d/projects.yaml
```diff
@@ -13,6 +13,8 @@
     name: telemetry-operator-multinode-autoscaling-tempest
     parent: telemetry-operator-multinode-autoscaling
     vars:
+      cifmw_test_operator_tempest_exclude_list: |
```
elfiesmelfie, line 16: You may need to add this to ci/vars-telemetry-autoscaling.yml instead.

vyzigold:
recheck
Looks like ceilometer pod restarted during the test (which is our current solution to get the latest set of keystone endpoints when they change after ceilometer is started)

jlarriba:
/lgtm

Downstream CI changes that will improve our testing capabilities

PR #394, opened 2024-05-23
Fix infinite ceilometer reconcile

vyzigold:
The endpointurls aren't sorted (they get returned in a different order every time). Unfortunatelly the hash function produces a different hash based on the order. This is fixed by sorting the urls


mrunge (approved):
Thank you, makes sense to me

jlarriba:
recheck

jlarriba:
recheck

PR #401, opened 2024-06-05
Get data from dataplane inventory secret

fao89:


controllers/metricstorage_controller.go
```diff
@@ -731,28 +693,26 @@ func getNodeExporterTargets(
 	if err != nil {
 		return []string{}, []string{}, err
 	}
-	nodeSetList, err := getNodeSetList(instance, helper)
+	inventorySecretList, err := getInventorySecretList(instance, helper)
```
jlarriba, line 696: We read the inventorySecretList during the controller reconcile, but we do not have any watch or any other mechanism that I can see that would trigger a new reconciler whenever this secret changes. In this case, how do we manage the quite common use case of adding or removing nodes?

fao89: I was thinking about adding a secret watcher, but I found this: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/controllers/metricstorage_controller.go#L917-L922 do you think it would already watch the inventory secrets or do we need a new watcher?

jlarriba: I am afraid no, that watcher only watches some Prometheus TLS related resources, specifically these ones: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/controllers/metricstorage_controller.go#L74

fao89:
recheck

jlarriba:
recheck 

It seems to be something legitimally wrong detected by the "default-telemetry" job but I am unable to see any kind of error log. Lets recheck and see if it fails again.

fao89:
recheck

jlarriba:
recheck

vyzigold:
@jlarriba The default-telemetry logs are in controller/ci-framework-data/logs/post_ctlplane_deploy_check_default_telemetry.log

It seems like it's some kind of issue with rabbit CRD not being on the cluster

https://logserver.rdoproject.org/01/401/22842c071a5ae4a7931846916c036258ec4aea13/github-check/telemetry-operator-multinode-default-telemetry/ef4d8f9/controller/ci-framework-data/logs/post_ctlplane_deploy_check_default_telemetry.log

jlarriba:
recheck

fao89:
recheck

PR #402, opened 2024-06-07
Define new autoscaling variables

vyzigold:
This should hopefully stabilize the autoscaling scenario downstream.


vyzigold:
recheck

PR #404, opened 2024-06-11
Switch "Enabled" to pointers

vyzigold:
It's easier to determine if the value has been set by the user with bool pointers instead of bools.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/839


csibbitt (approved):

vyzigold:
We need the Revert in O-O merged first.

vyzigold:
recheck

PR #410, opened 2024-06-13
Scrape RabbitMQ prometheus metrics

vyzigold:
This creates new ServiceMonitors, one per each RabbitMQ. RabbitMQ exposes its metric endpoint by default. The PR also changes the naming scheme of the ServiceMonitors from simply having one ServiceMonitor called `<instance-name>` to: `<instance-name>-<service-it-monitors-name>.<namespace>.svc`. So for example: `metric-storage-ceilometer-internal.openstack.svc` or `metric-storage-rabbitmq.openstack.svc`

### DIFF
--- a/controllers/metricstorage_controller.go
+++ b/controllers/metricstorage_controller.go
@@ -407,10 +407,10 @@ func (r *MetricStorageReconciler) reconcileNormal(
 	}
 	// ServiceMonitors for RabbitMQ monitoring
 	rabbitList := &rabbitmqv1.RabbitmqClusterList{}
-	listOpts := &client.ListOptions{
-		Namespace: instance.Namespace,
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.GetNamespace()),
 	}
-	err = r.Client.List(ctx, rabbitList, listOpts)
+	err = r.Client.List(ctx, rabbitList, listOpts...)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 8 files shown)


pkg/metricstorage/service_monitor.go
```diff
@@ -85,12 +86,15 @@ func ServiceMonitor(
 			},
 		},
 	}
+	if port != "" {
```
csibbitt, line 89: :+1:

controllers/metricstorage_controller.go
```diff
@@ -400,6 +405,73 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+		})
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if op != controllerutil.OperationResultNone {
+			Log.Info(fmt.Sprintf("Rabbit ServiceMonitor %s successfully changed - operation: %s", rabbitMonitor.Name, string(op)))
+		}
+	}
+	// Check that RabbitMQ monitor's RabbitMQs still exist
+	// Delete the ServiceMonitors, which don't have a RabbitMQ anymore
```
csibbitt, line 443: :+1:

controllers/metricstorage_controller.go
```diff
@@ -857,6 +929,33 @@ func (r *MetricStorageReconciler) SetupWithManager(ctx context.Context, mgr ctrl
...
 
+	rabbitmqWatchFn := func(ctx context.Context, o client.Object) []reconcile.Request {
+		result := []reconcile.Request{}
+
+		// get all metricstorage CRs
+		metricStorages := &telemetryv1.MetricStorageList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(o.GetNamespace()),
+		}
+		if err := r.Client.List(context.Background(), metricStorages, listOpts...); err != nil {
```
csibbitt, line 940: This might be for my own education rather than a request to change anything, but can you explain why the listOpts is in a different form here than in the other examples above?

- list of ListOption vs pointer to ListOption**s** -  on the surface they seem equivalent, just wondering if there was a motivation behind the inconsistency
-  `InNamespace()` instead of `Namespace:` - I did some reading but couldn't figure this one out

Also, why do you pass a fresh context to the List() call rather than propagate the one that the WatchFn is called with?

csibbitt (approved):
This LGTM but I do have questions

controllers/metricstorage_controller.go
vyzigold: The short answer is, that I just copied the prometheusServiceWatchFn just above this and adjusted it as I needed. But you have some interesting questions.

PR #411, opened 2024-06-14
Update hooks in Zuul jobs

elfiesmelfie:
There are two ways to define hooks in ci-framework: either as a list or as a single hook in a parameter.
The second option makes it easier to combine jobs, since each new hook is defined in its own var.


jlarriba:
/hold

lets wait for GA branch

PR #426, opened 2024-07-03
Removing RabbitMQ replace

vyzigold:
It's possible the replace isn't needed anymore. It was there to unblock CI back then. If we can't just remove it, we could at least update it to newer version.


vyzigold:
The CI passed, so it looks like the line isn't needed anymore. It was added after openstackversions was introduced and if I remember correctly, part of the issue was our dependency on dataplane-operator. The merge of openstack-operator and dataplane-operator might be the reason why we don't need this anymore.

jlarriba:
Could we please extend this PR to also add the other dependency change that Gibi realized it was not useful? Im referring to also remove the line https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/go.mod#L89 and let the CI run again.

PR #440, opened 2024-07-11
Consolidate duplicated RELATED_IMAGES

dprince:
This gets rid of 'make bundle' warnings both here and in the resulting openstack-operator variables we have for RELATED_IMAGES


vyzigold:
The aodh-api image here was used just because it includes httpd (we use httpd to termitate TLS). I see that in the openstackversion CR we have the "apacheImage", which I think would make more sense. What would be the way to use that image?

dprince:
After discussing with @jlarriba on slack we will go with slightly different approach here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/463

PR #442, opened 2024-07-12
[OSPRH-8508] Support ipv6 for metrics retrieval

jlarriba:


controllers/metricstorage_controller.go
```diff
@@ -412,78 +410,69 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+	svcScrapeList := &monv1alpha1.ScrapeConfigList{}
+	err = r.Client.List(ctx, svcScrapeList, listOpts...)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
-	for _, svcMonitor := range svcMonitorList.Items {
-		if svcMonitor.OwnerReferences == nil ||
-			len(svcMonitor.OwnerReferences) < 1 ||
-			svcMonitor.OwnerReferences[0].Name != instance.Name {
+	for _, scrapeConfig := range svcScrapeList.Items {
```
vyzigold, line 444: This loop was there to delete serviceMonitors for rabbitMQs, which don't exist anymore. With ServiceMonitors, each RabbitMQ required its own ServiceMonitor unfortunatelly, because I didn't find a selector for all of them.

With ScrapeConfigs we're basically the ones doing the service discovery. You could create a single ScrapeConfig for all rabbits. That way the you'd just need to update the ScrapeConfig's targets each reconciliation loop (which you already do) to keep them up to date.

vyzigold (changes requested):
I see that for some reason we worked on the same thing. This is my take from last week: https://github.com/vyzigold/telemetry-operator/commit/28ee41e5aa7debf40052c709398c6202dd1286cd

My biggest concern is (hence the request for changes), that I'm quite sure the rabbitmq dashboard won't work if we merge it like this. The RabbitMQ dashboard uses the "service" label to distinguish between individual rabbit instances. With ScrapeConfigs, there are no "service" labels added to the metrics.

Is this meant to get into GA? If not, how will updates work? What happens with the old ServiceMonitors? Should we delete them somewhere?

This will need changes to the kuttl tests as well.

controllers/metricstorage_controller.go
vyzigold: See how you could put all rabbits into a single ScrapeConfig: https://github.com/vyzigold/telemetry-operator/commit/28ee41e5aa7debf40052c709398c6202dd1286cd#diff-a4395c9fc4b59a814ac755d3a9faa9820081bc4300cd49f476b990defe28c1a2R503

jlarriba:
@vyzigold was already working on this, he will be submitting his own PR in replace of this one.

vyzigold:
https://github.com/openstack-k8s-operators/telemetry-operator/pull/446

PR #444, opened 2024-07-15
[OSPRH-8582] telemetry-operator must support empty template

jlarriba:


jlarriba:
/hold

I am not sure if this solves the issue, as I am unable to test it

jlarriba:
/unhold

I could test that it works

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/telemetry-operator
-  newTag: latest
+  newName: quay.io/jlarriba/telemetry-operator
+  newTag: v0.0.1
```
vyzigold, line 16:
```suggestion
  newName: quay.io/openstack-k8s-operators/telemetry-operator
  newTag: latest
```

api/v1beta1/metricstorage_webhook.go
```diff
@@ -55,8 +55,8 @@ func (spec *MetricStorageSpec) Default() {
 		// NOTE: If we want to enable dashboards in the future by default, set
 		//       it here like Alerting above
 	}
-	if spec.DataplaneNetwork == "" {
-		spec.DataplaneNetwork = "ctlplane"
+	if spec.DataplaneNetwork == nil || *spec.DataplaneNetwork == "" {
+		*spec.DataplaneNetwork = "ctlplane"
```
vyzigold, line 59:
```suggestion
		spec.DataplaneNetwork = ptr.To("ctlplane")
```
You'll need to import "k8s.io/utils/ptr" otherwise this will panic in case "spec.DataplaneNetwork == nil"

vyzigold: After a discussion and trying a lot of different things, we found out kubernetes won't allow the field to be nil as there isn't the "nullable" property in the CRD. Kubernetes will always set the field to some value other than nil see: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable

So initializing the pointer isn't required at least until we decide we want to add the "nullable" property.

jlarriba:
/test telemetry-operator-build-deploy

error: build error: Failed to push image: trying to reuse blob sha256:e394ea8406c7ed85ddc862f882ec77dcfd3863de3cb90fd8006238d521d22d44 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout

PR #445, opened 2024-07-15
Adding the bundle build and usage procedure in the README

jlarriba:


csibbitt (approved):
I haven't run this procedure myself yet, but it makes sense to read and seems like it would work

PR #446, opened 2024-07-15
[OSPRH-8406] Switch to using ScrapeConfigs

vyzigold:
We recently discovered issues with authentication, IPv6 and ServiceMonitors in STF. This PR is proactively switching to use ScrapeConfigs instead of ServiceMonitors. The functinality should be equivalent to before. Old ServiceMonitors owned by the MetricStorage controller are deleted.

There is a slight difference in the labels associated with the collected metrics.
 - The Node Exporter metrics are now missing the "job" label, which didn't seem useful and it follows how ceilometer and rabbit metrics are collected.
 - Ceilometer and RabbitMQ metrics don't have the "service" label anymore, because ScrapeConfigs don't have the information to create that label. Instead they now have the "instance" label.

The "instance" label is now used to differentiate between different Rabbit clusters in dashboards instead of the "service" label.

I used this opportunity to move the ScrapeConfig creation code into its own function, following the example of dashboard code.

### DIFF
--- a/pkg/metricstorage/scrape_config.go
+++ b/pkg/metricstorage/scrape_config.go
@@ -37,7 +37,10 @@ func ScrapeConfig(
 	var scrapeInterval monv1.Duration
 	if instance.Spec.MonitoringStack != nil && instance.Spec.MonitoringStack.ScrapeInterval != "" {
 		scrapeInterval = monv1.Duration(instance.Spec.MonitoringStack.ScrapeInterval)
-	} else if instance.Spec.CustomMonitoringStack != nil && *instance.Spec.CustomMonitoringStack.PrometheusConfig.ScrapeInterval != monv1.Duration("") {
+	} else if instance.Spec.CustomMonitoringStack != nil &&
+		instance.Spec.CustomMonitoringStack.PrometheusConfig != nil &&
+		instance.Spec.CustomMonitoringStack.PrometheusConfig.ScrapeInterval != nil &&
+		*instance.Spec.CustomMonitoringStack.PrometheusConfig.ScrapeInterval != monv1.Duration("") {
 		scrapeInterval = *instance.Spec.CustomMonitoringStack.PrometheusConfig.ScrapeInterval
 	} else {
 		scrapeInterval = telemetryv1.DefaultScrapeInterval


controllers/metricstorage_controller.go
```diff
@@ -366,161 +365,31 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
-			continue
-		}
-		rabbitmqExists := false
-		for _, rabbit := range rabbitList.Items {
-			if svcMonitor.Name == fmt.Sprintf("%s-%s.%s.svc", instance.Name, rabbit.Name, instance.Namespace) {
-				rabbitmqExists = true
-			}
-		}
-		if !rabbitmqExists {
-			err = r.Client.Delete(ctx, svcMonitor)
```
vyzigold, line 462: The deletion of ServiceMonitors (or ScrapeConfigs) for non-existing RabbitMQs isn't needed with ScrapeConfigs. Previously each instance of RabbitMQ would have its own ServiceMonitor, which would need to be deleted when the RabbitMQ instance was deleted. Now there is only one ScrapeConfig for all RabbitMQ instances. The targets inside the ScrapeConfig are updated each reconciliation loop, so it's always up to date.

controllers/metricstorage_controller.go
```diff
@@ -366,161 +365,31 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+	for _, monitor := range monitorList.Items {
+		if object.CheckOwnerRefExist(instance.ObjectMeta.UID, monitor.ObjectMeta.OwnerReferences) {
+			err = r.Client.Delete(ctx, monitor)
```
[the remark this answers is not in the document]
jlarriba: This should be on the reconcileUpdate func, however we are not yet using it.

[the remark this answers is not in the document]
vyzigold: I'll create the function and add it there then since I have the code ready. We can figure out how / when to call the function later I guess.

controllers/metricstorage_controller.go
```diff
@@ -366,43 +414,123 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
 	}
 
-	// ServiceMonitor for ceilometer monitoring
+	// ScrapeConfig for ceilometer monitoring
 	ceilometerServerName := fmt.Sprintf("%s-internal.%s.svc", ceilometer.ServiceName, instance.Namespace)
-	ceilometerMonitor := &monv1.ServiceMonitor{
+	ceilometerScrapeConfig := &monv1alpha1.ScrapeConfig{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-%s", instance.Name, ceilometerServerName),
+			Name:      fmt.Sprintf("%s-ceilometer", instance.Name),
```
jlarriba, line 509: I think we should be using telemetry.ServiceName here instead of instance.Name. The instance is called metric-storage, so the ScrapeConfigs get created as metric-storage-ceilometer and such. I think it is a much better idea to have them created as telemetry-ceilometer.

vyzigold: sure

PR #448, opened 2024-07-19
Revert "Remove memcached from autoscaling"

vyzigold:
This reverts commit bc526c54fc5c5809df48ad861ef285f7b91954f4.

Aodh needs to cache keystone tokens. In process caching is deprecated for a long time and we get a deprecation warning in the logs. This returns memcached support, which is used for this caching and it gets rid of the warning.

The warning:
```
[Tue Jul 02 13:51:22.466076 2024] [wsgi:error] [pid 14:tid 104] [remote 10.217.0.2:57004] 2024-07-02 13:51:22.465 14 WARNING keystonemiddleware.auth_token [-] Using the in-process token cache is deprecated as of the 4.2.0 release and may be removed in the 5.0.0 release or the 'O' development cycle. The in-process cache causes inconsistent results and high memory usage. When the feature is removed the auth_token middleware will not cache tokens by default which may result in performance issues. It is recommended to use  memcache for the auth_token token cache by setting the memcached_servers option.\x1b[00m  
```

rh-jira: https://issues.redhat.com/browse/OSPRH-8249

### DIFF
--- a/api/bases/telemetry.openstack.org_autoscalings.yaml
+++ b/api/bases/telemetry.openstack.org_autoscalings.yaml
@@ -42,6 +42,7 @@ spec:
                   databaseInstance: openstack
                   evaluatorImage: ""
                   listenerImage: ""
+                  memcachedInstance: memcached
                   notifierImage: ""
                   passwordSelector:
                     aodhService: AodhPassword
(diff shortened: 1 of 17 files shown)


vyzigold:
recheck

PR #458, opened 2024-08-02
Remove dataplane-operator references

rlandy:
Now that the dataplane-operator is combined and
not installed through a separate repo.


rlandy:
recheck

PR #470, opened 2024-08-19
Stop using Martin's images for autoscaling kuttl

vyzigold:
Images shouldn't matter for kuttl.


vyzigold:
recheck

Retry limit

vyzigold:
recheck

PR #474, opened 2024-09-03
Change how MetricStorage is copied in telemetry

vyzigold:
Chang the way the Telemetry controller copies MetricStorage spec from the Telemetry CR to a MetricStorage CR to match the rest of the code.

Fixes: https://issues.redhat.com/browse/OSPRH-9870

This will also prevent similar bugs in the future.


vyzigold:
/retest

Looks like some issue with rdo quay.

PR #479, opened 2024-09-18
Add Kepler's scrape endpoints to prometheus config

yadneshk:
Generate Kepler endpoints just like Node Exporter by appending corresponding ports to their services.
Also rename `getNodeExporterEndpoint` to `getMetricExporterEndpoint` since Kepler would also use it.


jlarriba:
Superseded by https://github.com/openstack-k8s-operators/telemetry-operator/pull/482

PR #484, opened 2024-09-20
Debugging kuttl failures

yadneshk:


jlarriba:
You are changing from "telemetry" to "telemetry-node-exporter" for the ScrapeConfig and the kuttl tests are complaining because those are no the same.

You have to change some of the kuttle tests, this one for example: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/tests/kuttl/suites/metricstorage/tests/01-assert.yaml#L79

There are more appearances of that ScrapeConfig that changed names, all of them need to be changed to the new name.

PR #494, opened 2024-09-27
[OSPRH-10380] Move dashboards to use go API

vyzigold:


jlarriba:
Please, add the jira issue that is being solved in the title of each PR with [] so Jira can make the binding.

I hope that with COO 0.3.1 we also have "Datasource" object to be able to add PrometheusDatasource object via API instead of just spitting plain yaml like we do here: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/pkg/metricstorage/dashboard_datasource.go#L46-L59

vyzigold:
I added the Jira number to the title. Regarding the Datasource, I also saw that piece of yaml. There is no "Datasource" type in OBO: https://github.com/rhobs/observability-operator/blob/main/pkg/apis/uiplugin/v1alpha1/types.go .

The issue with this yaml is, that it's not actually a CR. This is saved as a ConfigMap: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/controllers/metricstorage_controller.go#L720

PR #499, opened 2024-10-08
Add power monitoring dashboard

yadneshk:


jlarriba:
/hold

It has been detected that the presence of Kepler makes the VM boot much slower, see last comment on: https://issues.redhat.com/browse/OSPCIX-503. I am putting this on hold until we know what is going to happen with Kepler.

PR #509, opened 2024-10-14
Fix X-Forwarded-Proto in non-TLS deployments

stuggi:
In the httpd container configuration there was an issue that "X-Forwarded-Proto" header was setting "https" protocol always, no matter if TLS was enabled in the deployment or not. It caused issues in neutron server with pagination when TLS was disabled as client was doing request to the "http://" endoint and got in response "https://" links to "next" and "prev" pages.

This patch fixes this issue by setting correct protocol for both cases: TLS and non-TLS.

Related: [OSPRH-10632](https://issues.redhat.com//browse/OSPRH-10632)


stuggi:
based on https://github.com/openstack-k8s-operators/neutron-operator/pull/423

stuggi:
From discussion with @olliewalsh and @slawqo this should probably removed here because if, it should be set where the client terminates

PR #516, opened 2024-10-21
Create new secret to store ipmi configuration files

yadneshk:
Currently, ipmi agent uses compute agent's secret to load config files during deployment but ipmi agent only requires "hardware" pollster to be enabled to report metrics.

As compute and ipmi agents are deployed by different roles, segregating their config files into different secrets makes sense.

Similarly, on the compute nodes [1] takes care of storing config files into different directories.

[1] openstack-k8s-operators/edpm-ansible#793


jlarriba:
/hold

code freeze currently enforced

PR #517, opened 2024-10-22
[OSPRH-10800] Watch MetricStorage from Autoscaling

vyzigold:
MetricStorage is our main source for autoscaling metrics. We default to MetricStorage if other Prometheus instance isn't explicitly configured, so we need to ensure that Autoscaling is configured correctly to access this MetricStorage.

We previously decided to not enforce an order of deployment of telemetry resources, so it's pretty valid to first deploy Autoscaling and only after that deploy a MetricStorage. But if MetricStorage isn't deployed, Autoscaling doesn't know if it should configure Aodh to use TLS for Prometheus access or not. Without watching the MetricStorage resources, Autoscaling doesn't know if a MetricStorage CR is created or changed. So it's possible to do the following:
 - Enable Autoscaling (it gets automatically configured to not use TLS for Prometheus)
 - Enable MetricStorage with TLS enabled
 - Aodh can't access Prometheus because of TLS issues.

With this PR Autoscaling reacts to MetricStorage changes and so Aodh gets always correctly configured.

### DIFF
--- a/controllers/autoscaling_controller.go
+++ b/controllers/autoscaling_controller.go
@@ -788,6 +788,35 @@ func (r *AutoscalingReconciler) SetupWithManager(ctx context.Context, mgr ctrl.M
 		}
 		return nil
 	}
+	metricStorageFn := func(_ context.Context, o client.Object) []reconcile.Request {
+		result := []reconcile.Request{}
+
+		// get all autoscaling CRs
+		autoscalings := &telemetryv1.AutoscalingList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(o.GetNamespace()),
+		}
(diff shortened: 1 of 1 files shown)


mrunge:
The build fails with
```
error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: 
initializing source docker://registry.access.redhat.com/ubi9/go-toolset:1.21: 
pinging container registry registry.access.redhat.com: Get "https://registry.access.redhat.com/v2/": 
dial tcp: lookup registry.access.redhat.com on 172.30.0.10:53: server misbehaving
```
This is an infrastructure issue and not related to this patch

vyzigold:
recheck - let's see if it works now.

vyzigold:
/retest
I think those tests failed for the same reason

PR #523, opened 2024-11-01
Log at ERROR level for SeverityError Conditions

csibbitt:
I found several cases where a reconcile error occurs, but the log message doesn't stand out because it was at INFO level. Most cases in this code use log.Error when the resource status is being marked unready w/ condition.SeverityError, so this makes it consistent.


jlarriba:
These are informative logs, not errors of the operator per se, thats why they are Info messages.

PR #524, opened 2024-11-06
Disable tempest tests where not required

vyzigold:
Some of our jobs use ansible tests instead of tempest. This disables tempest on those jobs.


vyzigold:
recheck

zuul.d/projects.yaml
```diff
@@ -13,6 +14,7 @@
     name: telemetry-operator-multinode-autoscaling-tempest
     parent: telemetry-operator-multinode-autoscaling
     vars:
+      cifmw_run_tests: true
```
elfiesmelfie, line 17: I prefer to add these into the ci/vars- files that get included by cifmw_extras, since that's where we usually add the job vars

elfiesmelfie (changes requested):

PR #528, opened 2024-11-13
[OSPRH-10956] Fix aodh api timeout default

vyzigold:
I noticed the default didn't appear on the rendered CRD


vyzigold:
recheck

PR #533, opened 2024-11-18
Instead of uuids, list out node hostnames in Kepler dashboard

yadneshk:
Update all queries to use hostname label instead.


jlarriba:
With node_exporter we are using fqdn now (https://github.com/openstack-k8s-operators/telemetry-operator/pull/535). Please, rebase on top of main and use fqdn instead of hostname for the Label to keep everything in sync.

yadneshk:
hostname

yadneshk:
Wait for https://github.com/openstack-k8s-operators/telemetry-operator/pull/543 to merge first

PR #550, opened 2024-11-27
Return nil at the end of createScrapeConfigs

vyzigold:
Currently the function returns the latest value of the "err" variable. This isn't what should be happening. If we want to return an error, we should do so in one of the `if err != nil {}` blocks. When the function gets to the end, it should be considered as successfuly executed.


vyzigold:
/retest
RDO timeout

vyzigold:
recheck

PR #551, opened 2024-11-27
[OSPRH-11073] Add mysqld_exporter

vyzigold:
This adds mysqld_exporter to the Ceilometer resource.

What I'd really appreciate some feedback:
 - Where exactly to put the status of mysqld_exporter in the CRD

Whats missing:
 - ScrapeConfig - followup PR
 - API, TLS and openstackversion support in openstack-operator - PR when this is merged


jlarriba (approved):
/lgtm

vyzigold:
/hold
This adds new api fields. AFAIK this should be backwards compatible, but we have a pre-commit related PR, which will enable automatic checking of CRD backwards compatibility coming soon in: https://github.com/openstack-k8s-operators/telemetry-operator/pull/549 . So lets wait until that PR merges so that this PR can go through that check.

vyzigold:
The crd-checker fails with the followin errors:
```
ERROR: "NoBools": crd/ceilometers.telemetry.openstack.org version/v1beta1 field/^.spec.mysqldExporterEnabled may not be a boolean
ERROR: "NoMaps": crd/ceilometers.telemetry.openstack.org version/v1beta1 field/^.status.mysqldExporterHash may not be a map 
```
Following the crd-checker suggestions would break already established patterns throughout all of the operators, while not providing any additional value. IMO we should just ignore the errors. This should be fixed when (and if) some larger oko effort to change all these values across all of the operators happens.

vyzigold:
/retest
`error: the server doesn't have a resource type "openstackversion"` Doesn't seem related.

vyzigold:
/retest
I believe this was supposed to fix the issue: https://github.com/openshift/release/pull/59733

PR #570, opened 2024-12-18
[ci] Make the fvt tests voting

elfiesmelfie:
These jobs have been running for a while. When they fail, the zuul status doesn't show. Make them voting so that the statuses are seen, and failures can be explicitly ignored, if the job is broken.

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -76,7 +76,7 @@
         - telemetry-operator-multinode-autoscaling-tempest
         - telemetry-operator-multinode-default-telemetry
         - functional-tests-on-osp18: &fvt_jobs_config
-            voting: false
+            voting: true
             required-projects:
               - name: infrawatch/feature-verification-tests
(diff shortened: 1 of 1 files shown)


vyzigold:
recheck
default-telemetry failed on retry_limit - unrelated
logging job failure is already fixed
metric-verification failed on deploying COO (but also doesn't seem related):
```
Post "https://api.crc.testing:6443/apis/operators.coreos.com/v1alpha1/namespaces/openshift-operators/subscriptions?fieldManager=kubectl-create&fieldValidation=Ignore": dial tcp 38.102.83.73:6443: connect: connection refused'
```

vyzigold:
There is another failure. I proposed a fix for it. Let's see if the job works now or if there is something else that needs to be fixed: https://github.com/openstack-k8s-operators/telemetry-operator/pull/572

vyzigold:
recheck All the fixes are merged. This should work now.

PR #572, opened 2024-12-19
[ci] Fix logging job

vyzigold:


vyzigold:
/hold
The logging job is optional right now (there is another PR proposed to make it required), so this PR would merge even if the job isn't fixed. Let's hold it, so we can manually check it works before merging.

jlarriba:
/retest

It seems like a timeout
"Error from server (NotFound): deployments.apps "rabbitmq-cluster-operator-controller-manager" not found"

vyzigold:
/unhold

The deploy-logging-dependencies passed successfuly, so I think this could be good enough to merge. There was a failure on the fvt part of the logging job. It tries to check that the dataplane-operator is running, but we don't deploy dataplane-operator anymore. A fix will be required in the fvt repo.
https://logserver.rdoproject.org/72/572/17f10d67e78bac294feca8274c68ce2013915778/github-check/functional-logging-tests-osp18/ff2a5dd/controller/ci-framework-data/logs/pre_deploy_deploy_logging_dependencies.log

PR #573, opened 2024-12-19
Allow attaching MetricStorage pod to additional networks

vkmc:
For network isolation outgoing traffic, service pods need to be attached to additional networks. To do this, the `k8s.v1.cni.cncf.io/networks` annotation needs to be set.

This change adds a networkAttachments parameter to the MetricStorage CRD.

Closes: https://issues.redhat.com/browse/OSPRH-11017


controllers/metricstorage_controller.go
```diff
@@ -122,6 +124,7 @@ func (r *MetricStorageReconciler) GetLogger(ctx context.Context) logr.Logger {
 //+kubebuilder:rbac:groups=rabbitmq.com,resources=rabbitmqclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=observability.openshift.io,resources=uiplugins,verbs=get;list;watch;create;patch
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
```
jlarriba, line 127: This permissions puzzle me. We are only granting read permissions on NADs, but I imagine that eventually, we will need to create our own "something" (network-attachment?), but we are not giving any write permissions on any object, so...

controllers/metricstorage_controller.go
```diff
@@ -500,6 +504,32 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.NetworkAttachmentsReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
+		}
+	}
+
+	// when job passed, mark NetworkAttachmentsReadyCondition ready
+	instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
```
jlarriba, line 531: As I commented above regarding read permissions, this is currently not doing anything except for querying for NAD.

I understand that this is a WIP, but I imagine that we are pending to do something more before setting the NetworkAttachmentsReadyCondition to true, right?

Sorry for the questions, just trying to understand the solution while it is underway :D

vkmc: All valid questions! As I understand how it works for other services, NADs are defined before hand (in our case, we want to attach to the ctlplane) and then the NAD is added as an annotation to the resource using the https://pkg.go.dev/github.com/openstack-k8s-operators/lib-common/modules/common@v0.5.1-0.20241216113837-d172b3ac0f4e/networkattachment#EnsureNetworksAnnotation method to format the annotation (something that is missing from this PR).

Right now the current PR is just exposing the config option and validating that a value is set in the reconciliation loop.

I'm trying to define the following
- How should I add this annotation to the MonitoringStack resource
- Whether set or not a default value (e.g. ctlplane) (all other services don't set a default)

stuggi: > All valid questions! As I understand how it works for other services, NADs are defined before hand (in our case, we want to attach to the ctlplane) and then the NAD is added as an annotation to the resource using the https://pkg.go.dev/github.com/openstack-k8s-operators/lib-common/modules/common@v0.5.1-0.20241216113837-d172b3ac0f4e/networkattachment#EnsureNetworksAnnotation method to format the annotation (something that is missing from this PR).
> 
> Right now the current PR is just exposing the config option and validating that a value is set in the reconciliation loop.
> 
> I'm trying to define the following
> 
> * How should I add this annotation to the MonitoringStack resource

like the other operators [1], when you check for the availability of the nad, you create an `nadList`. with this we have a func in lib-common to create the correct annotation for the nad [2]. you then just have to pass/merge the annotation with maybe others you set on any deployments which need it [3][4].

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderbackup_controller.go#L473
[2] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderbackup_controller.go#L501
[3] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderbackup_controller.go#L518
[4] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/pkg/cinderbackup/statefulset.go#L110

> * Whether set or not a default value (e.g. ctlplane) (all other services don't set a default)

I'd not set a default to be in sync with the other service operator behavior. I think its better to report an failure of a missing nad, then report a wrong nad is missing if a user names it different, but the default requested ctlplane.

controllers/metricstorage_controller.go
```diff
@@ -502,6 +507,100 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+			return ctrl.Result{RequeueAfter: telemetryv1.PauseBetweenWatchAttempts}, nil
+		}
+		prometheusNADPatch := metricstorage.PrometheusNAD(instance)
+		err = r.Client.Patch(context.Background(), &prometheusNADPatch, client.Apply, client.FieldOwner("telemetry-operator"))
+		if err != nil {
+			Log.Error(err, "Can't patch Prometheus resource")
+			return ctrl.Result{}, err
+		}
+		instance.Status.NetworkAttachments = networkAnnotations
+	} else if len(instance.Spec.NetworkAttachments) == 0 {
```
vyzigold, line 581: I think this will always in every reconciliation delete the Prometheus CR when there aren't NADs defined. I see this is inspired by the TLS patch code above. In the TLS code, if the CR is patched (the if branch), it's saved in the metricStorage status and then it's checked before the deletion, so Prometheus CR gets deleted only in cases, when it's actually needed. I think something like that is needed here.

vyzigold (changes requested):
otherwise except for the issues pointed by pre-commit, it looks good to me

api/v1beta1/metricstorage_types.go
```diff
@@ -113,7 +118,7 @@ type MetricStorageSpec struct {
 	//       See rh-jira: OSPRH-5177 and COO-44
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:tpe=spec
```
amoralej, line 121: typo? I gues this should be `type=spec`

api/v1beta1/metricstorage_types.go
```diff
@@ -127,10 +132,13 @@ type MetricStorageStatus struct {
 	// then the controller has not processed the latest changes injected by
 	// the openstack-operator in the top-level CR (e.g. the ContainerImage)
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// NetworkAttachments status of the Prometheus pods
+	NetworkAttachments map[string]string `json:"networkAttachments,omitempty"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="NetworkAttachments",type="string",JSONPath=".status.networkAttachments",description="NetworkAttachments"
```
amoralej, line 141: This is adding NetworkAttachment to the `oc get metricstorage` output and makes it messy:

```
$ oc get metricstorage metric-storage
NAME             NETWORKATTACHMENTS                                                                                                                                          STATUS   MESSAGE
metric-storage   {"k8s.v1.cni.cncf.io/networks":"[{\"name\":\"ctlplane\",\"namespace\":\"openstack\",\"interface\":\"ctlplane\",\"default-route\":[\"172.18.250.254\"]}]"}   True     Setup complete

```

Do we want this to be included in the global status output?

amoralej (commented):
It worked fine in my local env with isolated networks (where I can reproduce the issue).

I left have a couple of comments. Other than that:

- Could we add coverage in kuttl tests? I see we are setting dataplaneNetwork to ctlplane in https://github.com/openstack-k8s-operators/telemetry-operator/blob/69a81e91783f161fa850fa469c3372c965a0f02c/tests/kuttl/suites/metricstorage/tests/06-deploy-edpm-resources.yaml so I think we may also add ctlplane to networkAttachment there.
- Should we add a Watches on NetworkAttachmentDefinitions to the metricstorage controller?. Doing so, a metricstorage which has failed because nad does not exist would trigger a reconcile when new nads are created.

vkmc:
Thanks for testing the proposed PR in your environment and for your review @amoralej!

api/v1beta1/metricstorage_types.go
amoralej: Right, I missed that. I see it as a too many information for a global status output but consistency is good :) . No problem from my side.

amoralej:
>     * Should we add a Watches on NetworkAttachmentDefinitions to the metricstorage controller?. Doing so, a metricstorage which has failed because nad does not exist would trigger a reconcile when new nads are created.

I was curious and check if other operators are doing the Watches, and they are not. I guess, given that NADs is something that must be pre-created in the cluster before deploying the controlplane it's probably not needed to add more logic to keep monitoring them from openstack-operators. If it's not present, it must be some error and the reconcile can be restarted by modifying the spec.

vyzigold:
Kuttl tests are failing. @vkmc I believe you're missing the new nad status condition in the metric-storage condition list in the default suite here: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/tests/kuttl/suites/default/tests/01-assert.yaml#L186

tests/kuttl/suites/default/tests/01-assert.yaml
```diff
@@ -185,6 +185,8 @@ status:
       status: "True"
     - type: TLSInputReady
       status: "True"
+    - type: NetworkAttachments
```
vyzigold, line 188:
```suggestion
    - type: NetworkAttachmentsReady
```

vyzigold: According to the kuttl-tests and other jobs https://logserver.rdoproject.org/73/573/8e6f472660a2739715db5f3eb9116a3db0062715/github-check/functional-graphing-tests-osp18/a9263ae/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/metricstorages.telemetry.openstack.org/metric-storage.yaml

vyzigold:
The functional tests failed with;
```
Failed to download packages: Curl error (6): Couldn't resolve host name for https://mirrors.centos.org/metalink?repo=centos-appstream-9-stream&arch=x86_64&protocol=https,http [Could not resolve host: mirrors.centos.org]"
```
doesn't seem related to the PR

vyzigold:
RDO quay timeout

vyzigold (approved):
looks good and the CI is green

PR #548, opened 2024-11-26
Change test var name to match changes in cifmw

eshulman2:
Changed var name to match the changes in this patch: https://github.com/openstack-k8s-operators/ci-framework/pull/2374/commits

PR #576, opened 2025-01-06
Fix test vars to fix downstream tests

vyzigold:
Fixes changes from openstack-k8s-operators/telemetry-operator#548

This makes our downstream tests fail. We need to wait for the ciframework change to get downstream first. Let's get back to this in like a month and try again.

### DIFF
--- a/ci/vars-autoscaling-tempest.yml
+++ b/ci/vars-autoscaling-tempest.yml
@@ -6,7 +6,7 @@ cifmw_test_operator_tempest_container: openstack-tempest-all
 cifmw_test_operator_tempest_image_tag: 'current-podified'
 # This value is used to populate the `tempestconfRun` parameter of the Tempest CR: https://openstack-k8s-operators.github.io/test-operator/crds.html#tempest-custom-resource
 # https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/test_operator/defaults/main.yml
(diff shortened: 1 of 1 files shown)


paramite (approved):

ci/vars-autoscaling-tempest.yml
```diff
@@ -6,7 +6,7 @@ cifmw_test_operator_tempest_container: openstack-tempest-all
 cifmw_test_operator_tempest_image_tag: 'current-podified'
 # This value is used to populate the `tempestconfRun` parameter of the Tempest CR: https://openstack-k8s-operators.github.io/test-operator/crds.html#tempest-custom-resource
 # https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/test_operator/defaults/main.yml
-cifmw_test_operator_tempest_tempestconf_config:
+cifmw_tempest_tempestconf_config:
```
elfiesmelfie, line 9: You don't have to rename it again, you could define both variables.

Or define a tempest_conf var (with this value), and then set both the new and old var names to "{{ tempest_conf }}"

elfiesmelfie (changes requested):

ci/vars-autoscaling-tempest.yml
vyzigold: Thanks for the suggestion

PR #577, opened 2025-01-06
[OSPRH-12079] Verify mysqld_exporter image is set

vyzigold:
This checks that the mysqld_exporter image is set. The mysqld_exporter reconciliation ends with an error if it isn't set.


vyzigold:
When setting the image to an empty string, the CR looked like this (I had to run the operator without webhooks, so that they don't overwrite the value with the default URL): https://paste.opendev.org/show/bmVPdgrjKcPcM9AQC3gL/

As expected the operator then waited with further reconciliation until a change to the CR happened. When I filled in a correct URL for the image, the mysqld_exporter deployed correctly.

PR #580, opened 2025-01-11
Move memcached defaulting into *Core* Default()

dprince:
Only the *Core* validation functions get used via
the OpenStackControlplane webhooks.


abays (approved):
/lgtm

dprince:
RDO job failure still has memcachedInstance getting set properly so I think it is unrelated

dprince:
recheck

dprince:
not sure why tide didn't pick up that LGTM

PR #585, opened 2025-01-16
Refactor Ceilometer services' status objects

paramite:
This patch joins status objects and uses prefixed contitions instead and splits hash object for each service.

This change is required as it fixes malfunctioning KSMStatus and adds missing hashing of KSM configuration.`


paramite:
This is resubmit of https://github.com/openstack-k8s-operators/telemetry-operator/pull/561 because kuttl test CI job keeps failing with no reason.

PR #593, opened 2025-01-29
[OSPRH-13223] Expose prometheus connection information in a secret

jlarriba:


controllers/metricstorage_controller.go
```diff
@@ -510,6 +518,60 @@ func (r *MetricStorageReconciler) reconcileNormal(
...
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	if instance.Spec.PrometheusTLS.Enabled() {
+		tlsSecret := &corev1.Secret{
+			Data: map[string][]byte{
+				"ca_secret": []byte(fmt.Sprintf("cert-%s-prometheus-svc", telemetryv1.DefaultServiceName)),
+				"ca_key":    []byte("ca.crt"),
```
vyzigold, line 558: We shouldn't assume the name of the secret. AFAIK this can be configured by users somehow instead of using the ones generated by O-O.
```suggestion
				"ca_secret": []byte(instance.Spec.PrometheusTLS.SecretName),
				"ca_key":    []byte(tls.CaKey),
```

vyzigold: See https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/pkg/metricstorage/prometheus_tls_patch.go#L53 for what's passed to Prometheus

vyzigold (changes requested):
Otherwise once the CI is green, it looks good

jlarriba:
recheck

paramite:
Legit. Shouldn't this PR contain also usage of the secret values by autoscaling deployment as is stated in the jira ticket?

vyzigold:
> Legit. Shouldn't this PR contain also usage of the secret values by autoscaling deployment as is stated in the jira ticket?

The autoscaling part of the ticket was meant as a suggested future enhancement. The main goal was to create the secret, so that other services can use it. Watcher will be the first to use it, autoscaling and openstackclient are the 2 other candidates, but we aren't much in a hurry to do that.

jlarriba:
> Legit. Shouldn't this PR contain also usage of the secret values by autoscaling deployment as is stated in the jira ticket?

To add to what Jaromir said, I really wanted this to go into 18.0.5, while the autoscaling modification can perfectly go into FR2, no rush with that

PR #594, opened 2025-01-30
[DNM] collect and visualize application logs

karelyatin:
Related-Issue: [OSPRH-13136](https://issues.redhat.com//browse/OSPRH-13136)

### DIFF
--- a/ci/vars-logging.yml
+++ b/ci/vars-logging.yml
@@ -3,6 +3,9 @@ cifmw_run_tests: false
 pre_deploy_deploy_logging_dependencies:
   source: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/deploy-logging-dependencies.yml"
   type: playbook
+pre_deploy_create_coo_subscription:
+    source: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/create-coo-subscription-playbook.yaml"
+    type: playbook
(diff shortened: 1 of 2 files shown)


ci/vars-logging.yml
```diff
@@ -3,6 +3,9 @@ cifmw_run_tests: false
 pre_deploy_deploy_logging_dependencies:
   source: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/deploy-logging-dependencies.yml"
   type: playbook
+pre_deploy_create_coo_subscription:
```
vyzigold, line 6: I see this is DNM, but I was curious, so I took a look. Do you actually need COO? AFAIK logging should be independent of COO.

karelyatin: yes it's independent, but i wanted this to have the logs brows able over Openshift UI. Or you mean it's not needed even for UI case as well?

vyzigold: Hmm. I thought you didn't need COO, but looking at the CRDs, which come with COO, I can see the UIPlugin there https://github.com/rhobs/observability-operator/blob/main/deploy/crds/common/observability.openshift.io_uiplugins.yaml#L11 . I think @jlarriba may know the best.

vyzigold:
/hold
Let's put this on hold just so this doesn't get merged by accident. Especially with the FR2 code freeze. You can unhold once FR2 branched and once this gets out of "DNM"

jlarriba:
Closing this as it never went past the "DNM" state. Please reopen again if it still valid.

PR #600, opened 2025-02-04
Autoscaling should use the prometheus secret

jlarriba:
In this PR, the PrometheusHost and PrometheusPort are read from the secret. However, aodh still continues to keep using the combined-ca-bundle instead of the ca_cert and ca_key combination, as that bundle is already mounted in the pod, so it would make little sense to mount the same certificate twice on the same pod.

The ca_cert/ca_key combination is provided for just in case other services (like Watcher) do not want to rely on the combined-ca-bundle to making it easier the integration with external prometheus.

I also included a couple of cleanups/small fixes.

### DIFF
--- a/controllers/autoscaling_controller.go
+++ b/controllers/autoscaling_controller.go
@@ -19,6 +19,8 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"github.com/openstack-k8s-operators/telemetry-operator/pkg/metricstorage"
+	"strconv"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


controllers/autoscaling_controller.go
```diff
@@ -425,13 +427,30 @@ func (r *AutoscalingReconciler) reconcileNormal(
...
+		Name:      autoscaling.PrometheusEndpointSecret,
+		Namespace: instance.Namespace,
+	}, prometheusEndpointSecret)
+	if err != nil {
+		Log.Info("No Prometheus Endpoint Secret found")
+	}
+
 	if instance.Spec.PrometheusHost == "" {
-		instance.Status.PrometheusHost = fmt.Sprintf("%s-prometheus.%s.svc", telemetryv1.DefaultServiceName, instance.Namespace)
+		if prometheusEndpointSecret.Data != nil {
```
paramite, line 440: Just a nit: does this need to be inner conditional?

controllers/autoscaling_controller.go
```diff
@@ -425,13 +427,30 @@ func (r *AutoscalingReconciler) reconcileNormal(
...
+			instance.Status.PrometheusHost = string(prometheusEndpointSecret.Data[metricstorage.PrometheusHost])
+		}
 	} else {
 		instance.Status.PrometheusHost = instance.Spec.PrometheusHost
 	}
 	if instance.Spec.PrometheusPort == 0 {
-		instance.Status.PrometheusPort = telemetryv1.DefaultPrometheusPort
+		if prometheusEndpointSecret.Data != nil {
```
paramite, line 447: Same here. It's just that to me the nesting does not seem necessary for me.

paramite (approved):

controllers/autoscaling_controller.go
jlarriba: My idea here is to nest it due to the else. We want the else to be executing just when instance.Spec.PrometheusHost == "" irrespective of the other condition, but you made me realized that this is how Go evaluates the "&&" conditions in an if. If The first operand is false, the second never gets evaluated.

vyzigold:
@jlarriba Don't we want to watch the secret? Can there be changes to the secret to which the autoscaling should react? We're already watching metricStorage, but maybe now that we're taking the connection parameters from the secret, we want to watch that instead.

jlarriba:
> @jlarriba Don't we want to watch the secret? Can there be changes to the secret to which the autoscaling should react? We're already watching metricStorage, but maybe now that we're taking the connection parameters from the secret, we want to watch that instead.

At first glance I didn't think so, but it is true that we want to avoid any kind of race condition. Watch on the secret will come in a follow up PR.

PR #603, opened 2025-02-05
Remove omitempty *Enabled fields

vyzigold:
This will make the fields more visible to the users to make them aware that there is something to enable / disable. The defaults stay the same, so no behavior changes.


vyzigold:
The crd-schema-checker is complaining unfortunately:
```
ERROR: "NoNewRequiredFields": crd/ceilometers.telemetry.openstack.org version/v1beta1 field/^.spec.ksmEnabled is new and may not be required
ERROR: "NoNewRequiredFields": crd/ceilometers.telemetry.openstack.org version/v1beta1 field/^.spec.mysqldExporterEnabled is new and may not be required
```

PR #608, opened 2025-02-11
[Dashboard] Update Legend for network traffic

mgirgisf:


vyzigold:
recheck

The 2 failures are just the test-operator installing twice (instead of just once as it's supposed to). The issue is known and it's being worked on. It appears randomly in jobs where tempest tests are executed. There is a chance for this to pass with a few rechecks.

PR #604, opened 2025-02-05
[OSPRH-12835] default-telemetry remove fail tasks

vyzigold:
Makes the following changes to the ansible of default-telemetry

- Remove quotes around task names (purely cosmetic change)
- Remove fail tasks in favor of failed_when and assert tasks
- Add TEST prefix and placeholder IDs to testcases
- Change some of the task names to be more descriptive
- Add the control_plane_name variable to make testing and development easier

Other changes will follow in future PRs. These will include

- splitting the long playbook and creating a role out of it
- using the fvt ansible config and executing the reporting playbook

PR #609, opened 2025-02-11
[OSPRH-13955][d-t] Get rid of fail tasks

vyzigold:
Remove fail tasks to enable reporting to polarion in the future. Part 1/4 of the original PR https://github.com/openstack-k8s-operators/telemetry-operator/pull/604

### DIFF
--- a/ci/check-default-telemetry.yml
+++ b/ci/check-default-telemetry.yml
@@ -34,23 +34,23 @@
     - name: "Get telemetry-operator logs"
       ansible.builtin.import_tasks: "get-operator-logs.yml"
 
-    - name: "Check telemetry-operator logs for Errors"
+    - name: |
+        TEST Check telemetry-operator logs for Errors
+        RHOSO-123456
       ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


ci/check-default-telemetry.yml
```diff
@@ -34,23 +34,23 @@
...
       when: error_list | length != 0
 
-    - name: "Get telemetry-operator restart counts"
+    - name: |
+        TEST Get telemetry-operator restart counts
+        RHOSO-123457
       ansible.builtin.import_tasks: "get-operator-restart-counts.yml"
+      failed_when: restart_counts != [0, 0]
 
     - name: "Fail if telemetry-operator was restarted"
```
elfiesmelfie, line 55: IMHO, this needs to be renamed as well, since the block now contains no fail conditions, just the output for debugging purposes

elfiesmelfie: If there's a plan to follow-up, then I am okay with leaving it unchanged here.

elfiesmelfie (approved):

PR #610, opened 2025-02-11
[OSPRH-13955][d-t] Extract oscp name into a var

vyzigold:
Extract the openstack control plane name into a variable, which allows for easier change of the name in the job if it the default name ever changes. It also allows for easier override form the command line,  when running the playbook locally
Part 2/4 of the original PR https://github.com/openstack-k8s-operators/telemetry-operator/pull/604

### DIFF
--- a/ci/check-default-telemetry.yml
+++ b/ci/check-default-telemetry.yml
@@ -21,6 +21,7 @@
           ["Ready", 'Setup complete']
         ])
       }}
+    control_plane_name: controlplane
   tasks:
     - name: "Get telemetry condition"
       ansible.builtin.command:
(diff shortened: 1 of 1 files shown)


ci/check-default-telemetry.yml
```diff
@@ -21,6 +21,7 @@
           ["Ready", 'Setup complete']
         ])
       }}
+    control_plane_name: controlplane
```
elfiesmelfie, line 24: This is great. I remember getting caught by this before when the oscp name changed in cifmw jobs.

vyzigold: To me it was always annoying as the controlplane is called differently in a default install_yamls deployment then whats in the CI, so it was annoying to run the playbook locally. This way I can specify the name with -e

PR #613, opened 2025-02-12
[OSPRH-13955][d-t] Make tasks more readable

vyzigold:
- Make task names more descriptive, so we can easily see how far the playbook is
- Remove quotes around task names

Part 3/4 of the original PR https://github.com/openstack-k8s-operators/telemetry-operator/pull/604

### DIFF
--- a/ci/check-default-telemetry.yml
+++ b/ci/check-default-telemetry.yml
@@ -1,5 +1,5 @@
 ---
-- name: "Check telemetry-operator logs for errors"
+- name: Execute the default-telemetry scenario
   hosts: "{{ cifmw_target_hook_host | default('localhost')  }}"
   gather_facts: false
   environment:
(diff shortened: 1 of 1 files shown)


ci/check-default-telemetry.yml
```diff
@@ -32,163 +32,161 @@
...
+      register: operator_logs_previous
+      when: restart_counts != [0, 0]
 
-        - name: "Output logs for debugging purposes"
-          ansible.builtin.debug:
-            var: operator_logs_previous.stdout_lines
+    - name: Output logs of failed container for debugging purposes
+      ansible.builtin.debug:
+        var: operator_logs_previous.stdout_lines
       when: restart_counts != [0, 0]
```
elfiesmelfie, line 66: It might be nicer to use a block/rescue here.
That is, if that works with the ignore_errors: true

If no, a block that includes both these tasks also makes sense.

vyzigold: Thanks for the suggestion. I wasn't sure which option I liked more. I'm not finished with the default-telemetry, so I'll keep this in mind.

vyzigold:
RDO timeout

ci/check-default-telemetry.yml
vyzigold: I tried creating a small playbook for testing. Unfortunately block/rescue doesn't seem to work with ignore_errors, but I'll at least put back the block (without rescue).

PR #614, opened 2025-02-13
Consume Topology CR by reference

fmount:
This patch provides more granular control over `Pod` placement and scheduling through the `Topology` CR integration introduced in `infra-operator`.
In particular it provides:

- a new `API` parameter (`TopologyRef`) defined for each `Component` that allows to reference an existing `Topology` CR in the same namespace

- the operator logic that retrieves and processes the referenced `Topology` CR through the functions provided by `infra-operator`

- enhanced `StatefulSet` and deployment configuration that incorporates the processed `Topology`

- a set of kuttl tests to double check the topology propagation

Note that `webhooks` are in place to prevent referencing a `Topology` from a different namespace (which is not a supported scenario).

Jira: https://issues.redhat.com/browse/OSPRH-14064

### DIFF
--- a/api/bases/telemetry.openstack.org_autoscalings.yaml
+++ b/api/bases/telemetry.openstack.org_autoscalings.yaml
@@ -345,6 +345,24 @@ spec:
                           a pre-created bundle file
                         type: string
                     type: object
+                  topologyRef:
+                    description: |-
+                      TopologyRef to apply the Topology defined by the associated CR referenced
+                      by name
+                    properties:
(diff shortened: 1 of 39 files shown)


jlarriba (approved):
Thank you Francesco!

fmount:
@jlarriba just rebased

jlarriba:
/lgtm

I have reviewed the document and it seems a very intelligent approach that gives our customers flexibility with their control plane deployments.

PR #624, opened 2025-02-20
Add missing dependency project

danpawlik:
The functional-tests-on-osp18 job requires openstack-k8s-operators-content-provider job to be set, but it was not defined.

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -95,3 +95,4 @@
         - functional-logging-tests-osp18: *fvt_jobs_config
         - functional-graphing-tests-osp18: *fvt_jobs_config
         - functional-metric-verification-tests-osp18: *fvt_jobs_config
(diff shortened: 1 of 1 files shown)


danpawlik:
recheck

danpawlik:
recheck

zuul.d/projects.yaml
```diff
@@ -95,3 +95,4 @@
         - functional-logging-tests-osp18: *fvt_jobs_config
         - functional-graphing-tests-osp18: *fvt_jobs_config
         - functional-metric-verification-tests-osp18: *fvt_jobs_config
+        - openstack-k8s-operators-content-provider
```
elfiesmelfie, line 98: This should be included in the podified-multinode-edpm-pipeline template.

If it's not, then this is something new, and would be effecting most other jobs too.

jlarriba:
This is not needed

PR #626, opened 2025-02-24
exporters: add dataplane node exporter

atyronesmith:
The dataplane node exporter is a Prometheus exporter that collects networking metrics from OVS, OVN and DPDK.  The
following changes include generation of the scrage config and a minimal dashboard.

[Depends-upon](https://issues.redhat.com/browse/OSPRH-11765)

DNM until image is available: quay.io/openstack-k8s-operators/dataplane-node-exporter

### DIFF
--- a/api/v1beta1/telemetry_consts.go
+++ b/api/v1beta1/telemetry_consts.go
@@ -28,6 +28,8 @@ const (
 	DefaultNodeExporterPort = 9100
 	// DefaultPodmanExporterPort -
 	DefaultPodmanExporterPort = 9882
+	// DefaultDataplaneNodeExporterPort -
+	DefaultDataplaneNodeExporterPort = 9105
 	// DefaultScrapeInterval -
 	DefaultScrapeInterval = "30s"
 	// PauseBetweenWatchAttempts -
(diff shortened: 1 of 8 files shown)


jlarriba:
Thank you for the contribution Aaron!

However, Dataplane Node Exporter does not seem as a descriptive name for what this exporter does, so I would request to get the name changed to something more descriptive. "ovs-exporter" seems to be a much better, yet concise name for this.

What do you think?

api/v1beta1/telemetry_consts.go
```diff
@@ -28,6 +28,8 @@ const (
 	DefaultNodeExporterPort = 9100
 	// DefaultPodmanExporterPort -
 	DefaultPodmanExporterPort = 9882
+	// DefaultDataplaneNodeExporterPort -
+	DefaultDataplanetNodeExporterPort = 9101
```
jlarriba, line 32: Due to incoming changes of ceilometer using this port, I would request that this be changed to 9105 if possible.

atyronesmith:
> Thank you for the contribution Aaron!
> 
> However, Dataplane Node Exporter does not seem as a descriptive name for what this exporter does, so I would request to get the name changed to something more descriptive. "ovs-exporter" seems to be a much better, yet concise name for this.
> 
> What do you think?

Hi.  The current version of the exporter collects data from both OVS and OVN.  In a future version, the exporter will include metrics for SR-IOV directly from the NICs.  The name "dataplane" was chosen to signify an exporter to collect metrics for network/data traffic as it flows through the dataplane.  We would prefer to keep dataplane, okay?

PR #633, opened 2025-03-05
[ci] Provide a zuul status to github when irrelevant files are changed

elfiesmelfie:
Adds the no-op job from feature-verificaton-tests, which will run and provide a passing status from zuul.
This is needed because github expects a status from zuul before it can mark a PR as passing/failing CI.

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -116,4 +116,5 @@
         - functional-logging-tests-osp18: *fvt_jobs_config
         - functional-graphing-tests-osp18: *fvt_jobs_config
         - functional-metric-verification-tests-osp18: *fvt_jobs_config
-        - feature-verification-tests-noop
+        - feature-verification-tests-noop:
+            files: *irrelevant_files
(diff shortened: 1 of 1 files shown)


zuul.d/projects.yaml
```diff
@@ -95,3 +95,6 @@
         - functional-logging-tests-osp18: *fvt_jobs_config
         - functional-graphing-tests-osp18: *fvt_jobs_config
         - functional-metric-verification-tests-osp18: *fvt_jobs_config
+        - feature-verification-tests-noop:
+            files: &irrelevant_files
+              - OWNERS*
```
elfiesmelfie, line 100: This works for Zuul giving _some_ response, but prow needs to be updated to similarly ignore some of these files

elfiesmelfie: Nevermind, it looks like prow just doesn't do anything on draft PRs

jlarriba:
/retest

`Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout`

PR #642, opened 2025-03-12
[ci] Enable tempest smoke tests

elfiesmelfie:


ci/vars-autoscaling-tempest.yml
```diff
@@ -21,8 +21,10 @@ tempest_conf:
 cifmw_tempest_tempestconf_config: "{{ tempest_conf }}"
 cifmw_test_operator_tempest_tempestconf_config: "{{ tempest_conf }}"
 cifmw_test_operator_tempest_include_list: |
+  \[.*\bsmoke\b.*\]
   telemetry_tempest_plugin.scenario
   telemetry_tempest_plugin.aodh
+# Should this tempest test be excluded?
```
vyzigold, line 27: It's excluded, because this particular job with the current configuration uses the original antelope containers. So no prometheus support and thus no autoscaling.

elfiesmelfie: Wait, is that the only autoscaling tempest test? 
Perhaps we should test the master containers in the telemetry-operator-multinode-autoscaling-tempest job.

Or run tempest in the autoscaling-fvt jobs, which uses the master containers

PR #645, opened 2025-03-13
Reconfigure ceilometer-compute agents to enable prom-exporter

mgirgisf:
This PR :
1- Disable RabbitMQ notification in ceilometer-agent-compute and enable prometheus exporter.
2- Add Scrape Config in prometheus configuration


templates/ceilometercompute/config/ceilometer.conf
```diff
@@ -27,12 +26,13 @@ rpc_address_prefix=
 notify_address_prefix=
 
 [oslo_messaging_notifications]
-driver=messagingv2
+driver=noop
```
jlarriba, line 29: With the enable_prometheus_exporter modification, we also added a simpler way to disable notifications. If you leave this as messagingv2 but change your polling section to:

[polling]
enable_notifications=False
heartbeat_socket_dir=/var/lib/ceilometer
enable_prometheus_exporter=true
prometheus_listen_addresses='0.0.0.0:9101'

that would do

mgirgisf:
recheck

templates/ceilometercentral/config/ceilometer.conf
```diff
@@ -60,4 +60,4 @@ telemetry_secret=eQ5qb0yysfJ8lx82Vl061vSyY
 [rgw_admin_credentials]
 [rgw_client]
 [service_types]
-[vmware]
+[vmware]
```
paramite, line 63: You gonna need to put this to original state to pass CI IMO.

templates/ceilometercentral/config/pipeline.yaml
```diff
@@ -9,4 +9,4 @@ sources:
 sinks:
   - name: tcp_sink
     publishers:
-      - tcp://127.0.0.1:3060
+      - tcp://127.0.0.1:3060
```
paramite, line 12: Ditto

mgirgisf:
recheck

mgirgisf:
recheck

mgirgisf:
recheck

mgirgisf:
recheck

jlarriba:
> I know there is some ongoing work by @jlarriba to get metrics from notifications in the future. I'm not 100% sure how that'll work, but with that in mind I don't know if we should disable RabbitMQ for Ceilometer. Juan what do you think?

You are absolutely right regarding the ongoing work, but this is disabling Notifications for ceilometer-compute-agent, which will not be publishing notifications anymore due to being directly scraped from Prometheus.

The proper RabbitMQ configuration for getting metrics from notifications goes into ceilometer-notification-agent in the control plane, which uses a different configuration from what Marihan is modifying.

api/v1beta1/telemetry_consts.go
```diff
@@ -30,6 +30,8 @@ const (
 	DefaultPodmanExporterPort = 9882
 	// DefaultDataplaneNodeExporterPort -
 	DefaultOpenStackNetworkExporterPort = 9105
+	// DefaultNodeExporterPort -
+	DefaultCeilometerPromExporterPort = 9101
```
jlarriba, line 34: Call this _DefaultCeilometerComputePromExporterPort_, to avoid confusions and to let space in the future in case we could need a _DefaultCeilometerCentralPromExporterPort_

tests/kuttl/suites/metricstorage/tests/01-assert.yaml
```diff
@@ -85,6 +85,18 @@ spec:
 ---
 apiVersion: monitoring.rhobs/v1alpha1
 kind: ScrapeConfig
+metadata:
+  labels:
+    service: metricStorage
+  name: telemetry-ceilometer-prom-exporter
```
jlarriba, line 91: This and the next one should change accordingly

mgirgisf:
recheck

PR #652, opened 2025-03-19
Drop power monitoring job since that is already being deployed by fvt

elfiesmelfie:
Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/242


elfiesmelfie:
recheck

PR #654, opened 2025-03-20
[18.0-FR2] Add overrides for required-projects

sdatko:
There are jobs that require projects that do not contain 18.0-fr2 branch
– for those, the override-checkout is needed so they can properly run in Zuul.

This commit contains workaround that may be reverted once the global solution is available.


sdatko:
recheck

sdatko:
recheck

PR #659, opened 2025-03-25
Enable prom-exporter with TLs

mgirgisf:

PR #678, opened 2025-05-06
Enable prom-exporter in ceilometer-ipmi

mgirgisf:
1- Disable RabbitMq  in ceilometer-ipmi 
2- Enable prom-exporter in ceilometer-ipmi 
3- Add scrapeconfig for the new port (9102)

### DIFF
--- a/api/v1beta1/telemetry_consts.go
+++ b/api/v1beta1/telemetry_consts.go
@@ -32,6 +32,8 @@ const (
 	DefaultOpenStackNetworkExporterPort = 9105
 	// DefaultCeilometerComputePromExporterPort -
 	DefaultCeilometerComputePromExporterPort = 9101
+	// DefaultCeilometerIpmiPromExporterPort -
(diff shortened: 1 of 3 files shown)


mgirgisf:
recheck

mgirgisf:
recheck

jlarriba:
> Some notes / questions. This will need TLS, so I guess something similar to #659 . But I guess that could be a followup PR.
> 
> I know there were some firewall issues with the NodeExporter port in the past. I think @jlarriba worked on fixing that. I wonder if there are PRs to modify the firewall rules to allow accepting traffic on ports 9102 (9101 for the compute agent)?
> 
> Otherwise, if we agree the TLS part would be a followup PR, this looks good.

Very well seen!

However, it has been already taken in account: https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/templates/firewall.yaml.j2

PR #684, opened 2025-05-12
Don't assume dns_search_domains item length

paramite:
This patch removes assumption that above mentioned item length at least 8 characters.

Fixes: [OSPRH-16162](https://issues.redhat.com//browse/OSPRH-16162)


vyzigold:
thank you

PR #693, opened 2025-06-04
[DNM] Try deploying autoscaling job with ceph

vyzigold:


vyzigold:
recheck

vyzigold:
recheck retry_limit

vyzigold:
Note to whoever is wondering about the future of the PR. As far as I can tell, this works pretty nicely, but I'd like to wait a little before merging. What should happen next:
- I'll do a bit of cleanup here
- I'll add a fvt PR, which uses this as depends-on and implements some actual tests
- Then I'll configure this job to run as part of a periodic pipeline

The plan can change slightly as I plan to coordinate with the rest of the team on details.

mgirgisf (approved):

PR #695, opened 2025-06-04
[ci] Add heat tempest tests to autoscaling-tempest job

elfiesmelfie:
Autoscaling feature rely on heat working. Adding the heat-tempest-plugin would help to ensure this dependency is working as expected.


vyzigold:
Having a quick look. Some of the heat tests failed. This one immediately strikes my eye: `heat_tempest_plugin.tests.scenario.test_aodh_alarm.AodhAlarmTest`. I'm pretty sure heat tries to find gnocchi. So there either needs to be some configuration value to tell it that gnocchi isn't available and to automatically skip these tests (we have something like this in the telemetry-tempest-plugin) or tests requiring gnocchi will need to be manually excluded.

vyzigold:
There is a long exclude list in the heat-operator tempest job. Maybe that could be the starting point for what to exclude: https://github.com/openstack-k8s-operators/heat-operator/blob/main/zuul.d/jobs.yaml#L29

ci/vars-autoscaling-tempest.yml
```diff
@@ -27,8 +27,14 @@ cifmw_test_operator_tempest_include_list: |
   ^tempest.*\[.*\bsmoke\b.*\]
   telemetry_tempest_plugin.scenario
   telemetry_tempest_plugin.aodh
+  ^heat_tempest_plugin.*
 cifmw_test_operator_tempest_exclude_list: |
   telemetry_tempest_plugin.scenario.test_telemetry_integration_prometheus.PrometheusGabbiTest.test_autoscaling
+  heat_tempest_plugin.tests.functional.test_os_wait_condition.OSWaitCondition
+  heat_tempest_plugin.tests.scenario.test_aodh_alarm.AodhAlarmTest
```
elfiesmelfie, line 34: The error for this is that the metric endpoint is not registered in keystone.

elfiesmelfie:
recheck

elfiesmelfie:
recheck

ci/vars-autoscaling-tempest.yml
elfiesmelfie: Should we try to fix this? Do we expect the metrics endpoint to be in keystone?

elfiesmelfie:
recheck

ci/vars-autoscaling-tempest.yml
vyzigold: No. "metric" endpoint means gnocchi.

elfiesmelfie:
recheck

elfiesmelfie:
recheck

PR #696, opened 2025-06-05
OSPRH-16753: Create an independent vars for graphing's deployment

MiguelCarpio:
**What does this PR do?**

Create an independent vars for graphing's deployment

**Why do we need it?**

For independent graphing's deployment that will allow us to run specific graphing tests on it.

### DIFF
--- a/ci/vars-graphing.yml
+++ b/ci/vars-graphing.yml
@@ -15,8 +15,6 @@ cifmw_edpm_prepare_kustomizations:
         metadata:
           name: unused
         spec:
-          heat:
-            enabled: true
           telemetry:
             enabled: true
             template:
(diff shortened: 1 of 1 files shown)


vyzigold:
the autoscaling failure is unrelated to the change, but it's interesting. What we're trying to do in the test is to create a stack, which adds a VM anytime the average load for all VMs in the stack is above 50%. I can see the values of the CPU load for each minute in the log. We start with 1 VM, we create  load, it gets to 100% CPU and a second VM appears. Ideally the first VM should stay on 100% CPU and a third VM should appear, but instead it fell to 0% or 1% (I can't distinguish which VM is which from the logs) even though the load on the first VM should still be the same. This is a little concerning and it's something we should keep an eye on going forwards. This is also the first failure like that I've seen. I'll keep an eye out for that as the CI Watcher this week.

vyzigold:
recheck

MiguelCarpio:
The `dashboardEnabled` is set in the autoscaling kustomization. https://github.com/openstack-k8s-operators/telemetry-operator/pull/698

PR #698, opened 2025-06-06
[ci] Extend the kustomizations in logging to include autoscaling

elfiesmelfie:
The cifmw_edpm_prepare_kustomizations var defined in vars-autoscaling gets replaced by the one in logging. This means that autoscaling doesn't get enabled.
For now, I have copied the various kustomizations into the same var in the logging file.
Going forward, this should be updated to copy the kustomizations into the artifacts/manifests/kustomizations/controlplane dir in separate hooks, so that they can be enabled/disabled independantly.


ci/create-coo-subscription-playbook.yaml
```diff
@@ -8,3 +8,47 @@
   tasks:
     - name: Import tasks for creating COO subscription
       ansible.builtin.include_tasks: "create-coo-subscription.yaml"
+
+    - name: Copy controlplane kustomization
```
vyzigold, line 12: This destroys the default-telemetry scenario, which relies on enabling / disabling parts of telemetry on its own as it goes

vyzigold: Also it'd be good to rename the file if it's doing more than just creating the COO sub.

vyzigold (changes requested):
As written in the PR description, I also think we should be creating the kustomization files, so this is going in the right direction, we just need to fix the issue with default-telemetry.

ci/create-coo-subscription-playbook.yaml
vyzigold: I see I'm incorrect, the default-telemetry works, because it actually doesn't check the environment it starts with. But I still think we shouldn't apply the kustomization for that job. One of the first steps in default-telemetry are:

1. Start without autoscaling and metricstorage enabled
2. Check that telemetry-operator works
3. Enable autoscaling and metricstorage
4. Check that everything went smooth and no errors are logged in telemetry-operator.

vyzigold: I left a comment on #700 . I think once that merges, then we could merge this one as well.

elfiesmelfie: I've rebased this branch against #700, so I don't have to do it later if my latest changes work and #700 merges shortly.

elfiesmelfie:
recheck

PR #701, opened 2025-06-13
Downgrade mysqld-exporter version

vyzigold:
Downgrade mysqld-exporter default image to 0.15.1 which is the last buildable version with go 1.21.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1487


vyzigold:
Looking at the CI job, it's executing the mysqld-exporter tests and it's passing. Manually checking the list of metrics in the CI against the list of metrics we're interested in, I see we have all of them, so we should be good to merge this.

PR #704, opened 2025-06-23
Add rbac-proxy container to metric-storage-prometheus pod

mgirgisf:


mgirgisf:
recheck

pkg/metricstorage/prometheus_proxy_patch.go
```diff
@@ -0,0 +1,82 @@
...
+			"kind":       "Prometheus",
+			"metadata": map[string]interface{}{
+				"name":      instance.Name,
+				"namespace": instance.Namespace,
+			},
+			"spec": map[string]interface{}{
+				"containers": []interface{}{
+					map[string]interface{}{
+						"name":  "kube-rbac-proxy",
+						"image": "quay.io/openstack-k8s-operators/kube-rbac-proxy:v0.16.0",
```
vyzigold, line 38: This line is an issue for merging it now. Any image we deploy needs to be configurable as we need to use upstream images in upstream (which is what's being done here), but we need to be able to use downstream images in downstream. For example, customers deploying in a disconnected environment will pre-download all required images and use those for the deployment. Those environments typically don't have direct access to the outside internet. In that case the `quay.io/openstack-k8s-operators/kube-rbac-proxy:v0.16.0` won't be accessible and the Prometheus pods will fail.

controllers/metricstorage_controller.go
```diff
@@ -417,6 +417,14 @@ func (r *MetricStorageReconciler) reconcileNormal(
 		return ctrl.Result{}, err
 	}
 
+	// Patch Prometheus service to add Proxy container
```
vyzigold, line 420: One of the proposals I made was to wrap this in `if kube_rbac_proxy_enabled` and set that to false for now. But looking at the PR, this would result in not deploying the container, which would in turn result in the kuttl-tests failing.

pkg/metricstorage/prometheus_proxy_patch.go
```diff
@@ -0,0 +1,82 @@
...
+	if tlsEnabled {
+		args = append(args,
+			"--upstream=https://metric-storage-prometheus.openstack.svc:9090/",
+			"--upstream-ca-file=/etc/tls/internal-ca-bundle.pem",
+		)
+	} else {
+		args = append(args, "--upstream=http://metric-storage-prometheus.openstack.svc:9090/")
+	}
+
+	prom := &unstructured.Unstructured{
```
vyzigold, line 26: Ideally we should use the golang types instead of the unstructured.Unstructured if we can like here: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/pkg/metricstorage/prometheus_tls_patch.go#L31 . But we could merge it and rewrite it later, we've done that previously.

vyzigold (changes requested):
As I see it, one of these things is required:

- Not merge this now. This is unfortunate after the amount of work and time went into this and it risks major merge conflicts once we get back to this (although the metricstorage part of the operator seems to be quite stable)
- We need to make the kube-rbac-proxy configurable. That would involve changing the metricstorage API to add an image field. Like this https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/api/v1beta1/ceilometer_types.go#L62 . We'd also need a change in the openstack-operator to set the image https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/telemetry.go#L341 .
- Wrap the code in metricstorage_controller.go in an if, so that this doesn't get executed. We'd also need to do something with the change in the prometheus_service_patch.go and delete all the changes to the kuttl-tests. This would  mean a lot of work with getting the kuttl-tests right again once we pick it up again.

PR #709, opened 2025-06-25
Add checks to prevent potential panics

vyzigold:
- Use two-return type assertions to check we have correct type
- Check for a variable not being nil before dereferencing
- Check for map item existence before accessing
- Fix log message grammar

Even thought I don't expect these issues to cause problems, unless there is a change in how inventory secrets look, I think this is still an improvement, which doesn't hurt anything. As far as I see, this doesn't change any behavior, which was also confirmed by my local testing. All scrapeconfigs were created correctly, compute node IPs were retrieved like before and I haven't seen any error logs in the operator logs.

Generated-By: cursor with claude-4-sonnet model

### DIFF
--- a/controllers/metricstorage_controller.go
+++ b/controllers/metricstorage_controller.go
@@ -1131,11 +1131,21 @@ func getComputeNodesConnectionInfo(
 		if err != nil {
 			return []ConnectionInfo{}, err
 		}
-		nodeSetGroup := inventory.Groups[secret.Labels["openstackdataplanenodeset"]]
+		nodeSetName, exists := secret.Labels["openstackdataplanenodeset"]
+		if !exists {
+			continue // Skip this secret if it doesn't have the required label
+		}
+		nodeSetGroup, exists := inventory.Groups[nodeSetName]
+		if !exists {
+			continue // Skip if the group doesn't exist in inventory
+		}
 		containsTargetService := false
-		for _, svc := range nodeSetGroup.Vars["edpm_services"].([]interface{}) {
-			if svc.(string) == telemetryServiceName {
-				containsTargetService = true
+		if services, ok := nodeSetGroup.Vars["edpm_services"].([]interface{}); ok {
+			for _, svc := range services {
+				if svcStr, ok := svc.(string); ok && svcStr == telemetryServiceName {
+					containsTargetService = true
+					break
+				}
 			}
 		}
 		if !containsTargetService {
@@ -1209,7 +1219,11 @@ func getAddressFromIPSet(
 	item *ansible.Host,
 	helper *helper.Helper,
 ) (string, discoveryv1.AddressType) {
-	ansibleHost := item.Vars["ansible_host"].(string)
+	ansibleHost, ok := item.Vars["ansible_host"].(string)
+	if !ok {
+		helper.GetLogger().Info("ansible_host is not a string or is missing")
+		return "", ""
+	}
 	canonicalHostname, _ := getCanonicalHostname(item)
 	ctlplaneDNSDomain := ""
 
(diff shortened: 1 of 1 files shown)


controllers/metricstorage_controller.go
```diff
@@ -1131,11 +1131,21 @@ func getComputeNodesConnectionInfo(
...
 		containsTargetService := false
-		for _, svc := range nodeSetGroup.Vars["edpm_services"].([]interface{}) {
-			if svc.(string) == telemetryServiceName {
-				containsTargetService = true
+		if services, ok := nodeSetGroup.Vars["edpm_services"].([]interface{}); ok {
+			for _, svc := range services {
+				if svcStr, ok := svc.(string); ok && svcStr == telemetryServiceName {
+					containsTargetService = true
+					break
+				}
```
jlarriba, line 1148: This is the "less bad" change of the whole set. I dont like having breaks in the code, but the checks actually do something. However, the line 

`if svcStr, ok := svc.(string); ok && svcStr == telemetryServiceName {`

Seems to be written by a very intelligent engineer just to show how intelligent he is.

controllers/metricstorage_controller.go
```diff
@@ -1260,7 +1274,10 @@ func getAddressFromIPSet(
 }
 
 func getAddressFromAnsibleHost(item *ansible.Host) (string, discoveryv1.AddressType) {
-	ansibleHost := item.Vars["ansible_host"].(string)
+	ansibleHost, ok := item.Vars["ansible_host"].(string)
+	if !ok {
+		return "", ""
+	}
```
jlarriba, line 1280: The same as above but more straightforward. If ansible_host does not exist, it will return "" which will fail the net.ParseIP(ansibleHost) func call returning nil and that will make the func exit with "", "".

controllers/metricstorage_controller.go
```diff
@@ -1131,11 +1131,21 @@ func getComputeNodesConnectionInfo(
...
 		}
-		nodeSetGroup := inventory.Groups[secret.Labels["openstackdataplanenodeset"]]
+		nodeSetName, exists := secret.Labels["openstackdataplanenodeset"]
+		if !exists {
+			continue // Skip this secret if it doesn't have the required label
+		}
+		nodeSetGroup, exists := inventory.Groups[nodeSetName]
+		if !exists {
+			continue // Skip if the group doesn't exist in inventory
+		}
```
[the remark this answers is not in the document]
vyzigold: I don't agree. Try creating a secret with exactly these labels (so that the `openstackdataplanenodeset` label is missing):
```
    labels:
      inventory: "true"
      openstack.org/operator-name: dataplane
```

This would mean, that the nodeSetGroup is an empty map. Not nil and that later fails on this line with a panic: https://github.com/openstack-k8s-operators/telemetry-operator/blob/e947500937f61c2a7abd36e07559d2a3ee497010/controllers/metricstorage_controller.go#L1141

Or take a valid inventory secret, but change the `openstackdataplanenodeset` label value to something else. You'll get a panic on the same line again.

These issues can be prevented either by these 7 lines or by the check below.

controllers/metricstorage_controller.go
vyzigold: While I see the `if svcStr, ok:= svc.string(); ok` part as a very common pattern throughout a lot of the go code I read. On a second look, I agree that the && part is a little too intelligent :D . I'll split it up. Regarding the break. It doesn't need to be there, just a small optimization of the loop, which I thought wouldn't hurt.

controllers/metricstorage_controller.go
```diff
@@ -1209,7 +1219,11 @@ func getAddressFromIPSet(
 	item *ansible.Host,
 	helper *helper.Helper,
 ) (string, discoveryv1.AddressType) {
-	ansibleHost := item.Vars["ansible_host"].(string)
+	ansibleHost, ok := item.Vars["ansible_host"].(string)
+	if !ok {
+		helper.GetLogger().Info("ansible_host is not a string or is missing")
+		return "", ""
+	}
```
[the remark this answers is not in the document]
vyzigold: I set the ansibleHost to `True`. At which point the original line failed. Also if the ansibleHost doesn't exist, why continue with the function and do k8s API requests? Why not just return early?

PR #710, opened 2025-06-26
Use advance pooling driver for MTLS

lmiccini:
This commit reworks the keystone_authtoken section to account for the new approach taken in https://review.opendev.org/c/openstack/keystonemiddleware/+/953174 where we use the advance pooling driver for MTLS.


jlarriba:
recheck

vyzigold:
recheck

PR #718, opened 2025-07-08
Fix broken pre-commit check

vyzigold:
Fix the pre-commit check broken in
https://github.com/openstack-k8s-operators/telemetry-operator/pull/716


jlarriba:
https://github.com/openstack-k8s-operators/telemetry-operator/pull/716 did not complained about this pre-commit fix, I guess because the pre-commit checks are not done when the README is modified.

I believe we should re-enable pre-commit checks on README changes

PR #722, opened 2025-07-10
[compute] Update prometheus_listen_address to include ipv6

elfiesmelfie:
Relates: [OSPRH-18158](https://issues.redhat.com//browse/OSPRH-18158)

### DIFF
--- a/templates/ceilometercompute/config/ceilometer.conf
+++ b/templates/ceilometercompute/config/ceilometer.conf
@@ -37,7 +37,7 @@ prometheus_tls_keyfile = {{ .TlsKey }}
 {{- end }}
 enable_notifications=False
 enable_prometheus_exporter=true
-prometheus_listen_addresses='0.0.0.0:9101,[::]:9101'
+prometheus_listen_addresses=['0.0.0.0:9101', '[::]:9101']
 
 [publisher]
 telemetry_secret=eQ5qb0yysfJ8lx82Vl061vSyY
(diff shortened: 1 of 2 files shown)


templates/ceilometeripmi/config/ceilometer.conf
```diff
@@ -30,7 +30,7 @@ driver=noop
 heartbeat_socket_dir=/var/lib/ceilometer
 enable_notifications=False
 enable_prometheus_exporter=true
-prometheus_listen_addresses='0.0.0.0:9102'
+prometheus_listen_addresses=0.0.0.0:9102,[::]:9102
 {{- if .TLS }}
```
elfiesmelfie, line 34: @jlarriba Should there have been something like this if TLS block in the ceiometercompute/config/ceilometer.conf file as well?

templates/ceilometercompute/config/ceilometer.conf
```diff
@@ -37,7 +37,7 @@ prometheus_tls_keyfile = {{ .TlsKey }}
 {{- end }}
 enable_notifications=False
 enable_prometheus_exporter=true
-prometheus_listen_addresses='0.0.0.0:9101'
+prometheus_listen_addresses='0.0.0.0:9101,[::]:9101'
```
mgirgisf, line 40: I think .. this need to be added like that prometheus_listen_addresses = ['0.0.0.0:9101', '[::]:9101']

elfiesmelfie: Thanks!

PR #733, opened 2025-08-05
Update metricStorage NAD kuttl tests

vkmc:
Drop NAD config from deploy-edpm-resources

Add configure-nad-metricstorage test which covers the scenario in which TLS is enabled and a NAD is set for a metricStorage.

Related Bug: https://issues.redhat.com/browse/OSPRH-14663

### DIFF
--- a/tests/kuttl/suites/tls/tests/03-assert.yaml
+++ b/tests/kuttl/suites/tls/tests/03-assert.yaml
@@ -1,10 +1,10 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: prometheus-telemetry-kuttl-metricstorage-with-nad-0
+  name: prometheus-telemetry-kuttl-metricstorage-0
   ownerReferences:
   - kind: StatefulSet
-    name: prometheus-telemetry-kuttl-metricstorage-with-nad
+    name: prometheus-telemetry-kuttl-metricstorage
   annotations:
     k8s.v1.cni.cncf.io/networks: '[{"name":"ctlplane","namespace":"openstack","interface":"ctlplane"}]'
 spec:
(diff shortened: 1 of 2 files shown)


vyzigold:
I don't think this tests much. In my understanding of our kuttl tests (I'm not an expert, so I could be wrong here), you have numbered stages. Here you're starting a 3rd stage of the tls test. In most of the stages, you have some CRs you want to be applied in the test (which is what you're doing) and then you have <stage_number>-assert.yaml file, where you check that all of the modified resources / subresources look as you want (which is what I think is missing). I think you should add a 03-assert.yaml, where you'll check at least that the metricstorage you're modifying is 'Ready'. If there is any modifications done to the subresources, you could check that as well. For example that the created Prometheus CR has the TLS config options and the NAD annotation in podMetadata like here https://logserver.rdoproject.org/e39/rdoproject.org/e39fd55813c34d23923f0d2dd94e3b4f/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/prometheuses.monitoring.rhobs/metric-storage.yaml

I'm just a little worried this will pass even if the NAD + TLS doesn't work if we don't add an assert to the stage.

vyzigold:
Over here you have the telemetry operator logs https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_telemetry-operator/733/pull-ci-openstack-k8s-operators-telemetry-operator-main-telemetry-operator-build-deploy-kuttl/1955943876023816192/artifacts/telemetry-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-6dfab6238607585487c0a518d5f94c13a553c98cd9f4d091b3a147c27303f22f/namespaces/openstack-operators/pods/telemetry-operator-controller-manager-755f9766c-gcsf8/logs/manager-previous.log . Notice the timestamps in the diff and look for what's happening in the operator logs around the similar timestamps. You can see it complains with `network-attachment-definition ctlplane not found`. You probably want to create some dummy resources, so that it looks like the ctlplane NAD exists. Kind of something similar like Yadnesh has done here https://github.com/openstack-k8s-operators/telemetry-operator/blob/f9e60d64606b8d541784f9154349abb0e3d782e7/tests/kuttl/suites/metricstorage/tests/05-prepare-for-edpm-power-mon.yaml This basically simulates that there are compute nodes present.

PR #734, opened 2025-08-07
[ci] Use smoke param for tempest test configs

elfiesmelfie:
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3184


elfiesmelfie:
The smoketest flag worked, but now there are failing murano and octavia tests.
Fixing this is a case of updating the exclude list

elfiesmelfie:
recheck

PR #741, opened 2025-08-26
Add CloudKitty deployment to telemetry-operator

jlarriba:


pkg/cloudkittyproc/statefulset.go
```diff
@@ -0,0 +1,152 @@
...
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 5,
+	}
+
+	args := []string{"-c", ServiceCommand}
+	var probeCommand []string
+	livenessProbe.HTTPGet = &corev1.HTTPGetAction{
+		Port: intstr.FromInt(8080),
+	}
+	startupProbe.HTTPGet = livenessProbe.HTTPGet
```
vyzigold, line 67: I wonder if the probes for the proc pod are correct. I wasn't able to get everything fully running yet, so this may be valid. But do we expect the proc process to listen on an HTTP port? If not, then the probes should use something similar to https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/pkg/ceilometer/statefulset.go#L91

jlarriba: Yes indeed. @mgirgisf added an appropiate healthcheck script in here https://github.com/openstack-k8s-operators/telemetry-operator/commit/1f8910664963e580c99b5c32d0c616227d406cf3, but it is true it seems not be used as we are still accessing the port instead of the script. This should be fixed.

controllers/cloudkittyapi_controller.go
```diff
@@ -0,0 +1,1118 @@
...
+						Name:      cr.Name,
+					}
+					Log.Info(fmt.Sprintf("Secret %s and CR %s marked with label: %s", o.GetName(), cr.Name, l))
+
+					result = append(result, reconcile.Request{NamespacedName: name})
+				}
+			}
+		}
+
+		// Watch for changes to any CustomServiceConfigSecrets
```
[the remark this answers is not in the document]
jlarriba: Because all the code is heavily based on cinder-operator code to be as compatible as possible with other operators, and Cinder does this, no other real reason, I just wanted this to do exactly the same as Cinder behaviour.

api/v1beta1/telemetry_types.go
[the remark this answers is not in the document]
jlarriba: ack

pkg/cloudkittyproc/statefulset.go
```diff
@@ -0,0 +1,152 @@
...
+								RunAsUser: &cloudKittyUser,
+							},
+							Env:           env.MergeEnvs([]corev1.EnvVar{}, envVars),
+							VolumeMounts:  volumeMounts,
+							Resources:     instance.Spec.Resources,
+							LivenessProbe: livenessProbe,
+							StartupProbe:  startupProbe,
+						},
+						{
+							Name:    "probe",
```
vyzigold, line 122: We're running an additional 'probe' sidecar container here, which seems like it's trying to run the healthcheck script (which doesn't succeed for me anyway). I wonder, is this second container intentional? Wouldn't it be possible to run the healthcheck script as a Liveness / Startup probe of the main container? Similar to what we do for ceilometer for example.

controllers/cloudkittyapi_controller.go
```diff
@@ -0,0 +1,1118 @@
...
+			ServiceName:        ksSvc["name"],
+			ServiceDescription: ksSvc["desc"],
+			Enabled:            true,
+			ServiceUser:        instance.Spec.ServiceUser,
+			Secret:             instance.Spec.Secret,
+			PasswordSelector:   instance.Spec.PasswordSelectors.CloudKittyService,
+		}
+
+		ksSvcObj := keystonev1.NewKeystoneService(ksSvcSpec, instance.Namespace, serviceLabels, cloudkitty.NormalDuration)
+		ctrlResult, err := ksSvcObj.CreateOrPatch(ctx, helper)
```
vyzigold, line 611: This will add `openstack.org/cloudkittyapi` into the list of finalizers of the created keystoneservice object. But this object is required by the cloudkitty proc as well (cloudkitty proc is communicating with other services and with keystone as well, so it needs the keystone service to exist).

So if we want to be absolutely correct, we should somehow add a `openstack.org/cloudkittyproc` finalizer there as well. I don't think this is a big issue though, so I wouldn't mind merging without fixing this.

templates/cloudkitty/config/cloudkitty.conf
```diff
@@ -0,0 +1,87 @@
...
+auth_type = v3password
+{{- if .TLS }}
+cafile = {{ .CAFile }}
+{{- end }}
+
+[fetcher]
+backend = keystone
+
+[fetcher_keystone]
+auth_section = authinfos
```
vyzigold, line 31: ignore_rating_role = True

templates/cloudkitty/config/cloudkitty.conf
```diff
@@ -0,0 +1,87 @@
...
+[collect]
+period = 300
+wait_periods = 0
+metrics_conf = /etc/cloudkitty/metrics.yaml
+collector = prometheus
+scope_key = project
+
+[collector_prometheus]
+{{- if .TLS }}
+prometheus_url = https://{{ .PrometheusHost }}:{{ .PrometheusPort }}
```
vyzigold, line 49:
```suggestion
prometheus_url = https://{{ .PrometheusHost }}:{{ .PrometheusPort }}/api/v1
```

pkg/cloudkitty/storageinit.go
```diff
@@ -0,0 +1,109 @@
...
+	//       services as described in the upstream rolling upgrades process.
+	storageInitCommand = "/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start"
+)
+
+// StorageInitJob func
+func StorageInitJob(instance *telemetryv1.CloudKitty, labels map[string]string, annotations map[string]string) *batchv1.Job {
+	args := []string{"-c", storageInitCommand}
+
+	// create Volume and VolumeMounts
+	volumes := GetVolumes("cloudkitty")
```
vyzigold, line 43:
```suggestion
	volumes := GetVolumes(instance.Name)
```
I'll fix that in my PR

pkg/cloudkitty/volumes.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		{
+			Name:      "config-data",
+			MountPath: "/var/lib/openstack/config",
+			ReadOnly:  true,
+		},
+		{
+			Name:      "config-data",
+			MountPath: "/var/lib/kolla/config_files/config.json",
+			SubPath:   serviceName + "-config.json",
```
vyzigold, line 53: The filename is hardcoded based on the name in templates/cloudkitty/confg/cloudkitty-<proc/api>-config.json, so imo this should move into the child volumes.go with `SubPath: cloudkitty.ServiceName + "-<api or proc>-config.json"`

I'll fix that in my PR.

pkg/cloudkittyproc/statefulset.go
mgirgisf: I think we can do .. it was similar to cinder healthcheck .. so I only changed the logic but its better practice to define the script in the main container as a livenessProbe or startupProbe as its  logic is not that heavy anyway

pkg/cloudkittyproc/statefulset.go
mgirgisf: this conflict because we have a probe container running the healthcheck instead of running it inside the container itself .. so I'm going to remove that extra container and update the config

controllers/telemetry_controller.go
```diff
@@ -553,6 +564,89 @@ func (r TelemetryReconciler) reconcileLogging(ctx context.Context, instance *tel
...
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			telemetryv1.CloudKittyReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+
+	// Check the observed Generation and mirror the condition from the
+	// underlying resource reconciliation
+	autoObsGen, err := r.checkCloudKittyGeneration(instance)
```
vyzigold, line 620:
```suggestion
	ckObsGen, err := r.checkCloudKittyGeneration(instance)
```

controllers/cloudkitty_controller.go
```diff
@@ -0,0 +1,1391 @@
...
+	// TransportURL created and used by CloudKitty CRs.  Watch functions return a list of namespace-scoped
+	// CRs that then get fed  to the reconciler.  Hence, in this case, we need to know the name of the
+	// CloudKitty CR associated with the secret we are examining in the function.  We could parse the name
+	// out of the "%s-cloudkitty-transport" secret label, which would be faster than getting the list of
+	// the CloudKitty CRs and trying to match on each one.  The downside there, however, is that technically
+	// someone could randomly label a secret "something-cloudkitty-transport" where "something" actually
+	// matches the name of an existing CloudKitty CR.  In that case changes to that secret would trigger
+	// reconciliation for a CloudKitty CR that does not need it.
+	//
+	// TODO: We also need a watch func to monitor for changes to the secret referenced by CloudKitty.Spec.Secret
```
vyzigold, line 272: Also watch the PrometheusEndpoint secret. The secret could change for example when switching TLS on / off. Also it might take the metricStorage some time to turn TLS off (prometheus will always be deployed without TLS at first), which could result in CK picking up the non TLS endpoint secret and then not reacting to when TLS gets enabled. So watching it would be good.

jlarriba: I already added the code for watching the PrometheusEndpoint secret to the cloudkittyapi_controller and cloudkittyproc_controller

vyzigold:
Seems like the LokiStackSize broke the interaction between O-O and T-O, which escaped me in my local testing. I'll provide a fix shortly.

PR #746, opened 2025-09-02
Extend loki timeout

lnatapov:
Sometimes logging deployment fails and as a result it fails the job. It fails on timeout when deploying loki.
 it seems all the pod is up and running as expected ,so I assume we can update the ansible task instead of using oc wait to use retry and delay

### DIFF
--- a/ci/deploy-logging-dependencies/tasks/deploy-loki.yml
+++ b/ci/deploy-logging-dependencies/tasks/deploy-loki.yml
@@ -19,5 +19,10 @@
 
 - name: Wait for the lokistack to be ready
   ansible.builtin.command:
-    cmd: |
-      oc wait --timeout=500s --for condition=Ready=True --namespace=openshift-logging lokistacks logging-loki
+    cmd: >
+      oc get lokistack logging-loki -n openshift-logging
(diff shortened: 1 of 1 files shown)


vyzigold (approved):

ci/deploy-logging-dependencies/tasks/deploy-loki.yml
```diff
@@ -19,5 +19,10 @@
...
 - name: Wait for the lokistack to be ready
   ansible.builtin.command:
-    cmd: |
-      oc wait --timeout=500s --for condition=Ready=True --namespace=openshift-logging lokistacks logging-loki
+    cmd: >
+      oc get lokistack logging-loki -n openshift-logging
+      -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}'
+  register: loki_ready
+  until: loki_ready.stdout == "True"
+  retries: 40
```
vyzigold, line 27:
```suggestion
  retries: 40
```
Can you please remove that trailing space? It won't merge otherwise.

vyzigold (changes requested):

elfiesmelfie (approved):

PR #752, opened 2025-09-16
[CI] Add test prom_secret to default telemetry job

mgirgisf:
This change ensures that during the test, the metric-storage-prometheus-endpoint secret is properly validated and matches the expected structure. which can verifies that the Prometheus endpoint is correctly exposed
and also confirms that the CA certificate, secret reference, host, and port are correctly set for secure access to Prometheus.


mgirgisf:
recheck

elfiesmelfie (changes requested):
Please update the commit  message to add more details (motivation, etc)

PR #766, opened 2025-09-29
Add Alertmanager sample alerts

vkmc:
This change creates a docs/ directory and adds sample alerts to be used with the Alertmanager instance deployed (when enabled) by the Metric Storage

Closes: OSPRH-10846


vkmc:
recheck

PR #876, opened 2026-04-24
DNM Test mariadb-operator timeout increase

vyzigold:
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/450


vyzigold:
The timeout increase worked. It got rid of one of the issues we were seeing, but ultimately the job was still failing. The timeout increase could be merged in if it's seen as a good idea from the mariadb-operator side, but it doesn't help telemetry much. The real fix is being worked on in: https://github.com/infrawatch/feature-verification-tests/pull/366

PR #877, opened 2026-04-24
DNM Disable swift and barbican for autoscaling job

vyzigold:


vyzigold:
We can't just disable swift. It's used as a storage backend for glance. We may disable Barbican (it'd need testing), but we have a different approach to fix the CI. We're bumping the node we use.

PR #880, opened 2026-04-27
Improved sample alerts.

csibbitt:
This patch makes our sample alert queries match the better ones from our KB article[1].

* These alerts give more specific information about what component is not running
* I've tried them in my own setup and the results look correct to me
* Credit to @vkmc on the queries, I'm just making things the same

[1] https://access.redhat.com/articles/7132715

### DIFF
--- a/docs/sample-alerts/openstack-observability-services-status.yaml
+++ b/docs/sample-alerts/openstack-observability-services-status.yaml
@@ -12,7 +12,7 @@ spec:
       - alert: OpenStackServicesDownWarning
         expr: |
           (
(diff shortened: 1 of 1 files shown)


csibbitt:
## TESTING

```
[csibbitt@laptop telemetry-operator (⎈|openstack/api-crc-testing:6443/kubeadmin:openstack)]$ oc apply -f docs/sample-alerts/openstack-observability-services-status.yaml

prometheusrule.monitoring.rhobs/openstack-observability-services-status created
```
image

image

vkmc (approved):
Thanks for following up on this! This query is definitely more useful

PR #882, opened 2026-04-29
[OSPRH-29177] Add Zuul job and Ansible role for CADF audit logging

vyzigold:
Add an enable-audit-logging role that creates audit config secrets and applies kustomizations to enable CADF audit logging on cinder, glance, keystone, and neutron. Includes a Zuul job definition and ci-framework vars file for CI, plus a local apply mode for manual use.

### DIFF
--- a/ci/enable-audit-logging/defaults/main.yml
+++ b/ci/enable-audit-logging/defaults/main.yml
@@ -1,0 +1,5 @@
+enable_audit_logging_namespace: "{{ cifmw_openstack_namespace | default('openstack') }}"
+# Set to true to apply the kustomization against a live OSCP CR.
+# When false (default), the kustomization is only copied to the
+# ci-framework kustomizations directory for pre-deploy use.
+enable_audit_logging_local_apply: false
--- a/ci/enable-audit-logging/tasks/apply-local.yml
+++ b/ci/enable-audit-logging/tasks/apply-local.yml
@@ -1,0 +1,29 @@
+- name: Create temp directory for local kustomization
+  ansible.builtin.tempfile:
+    state: directory
+    prefix: audit-logging-
+  register: audit_kustomize_dir
+
+- name: Fetch current OSCP CR
+  ansible.builtin.shell: >
+    oc get openstackcontrolplane -o yaml
+    > {{ audit_kustomize_dir.path }}/oscp.yaml
+
+- name: Render kustomization to temp directory
+  ansible.builtin.template:
+    src: 90-kustomize-controlplane-audit-logging.yaml.j2
+    dest: "{{ audit_kustomize_dir.path }}/kustomization.yaml"
+
+- name: Add OSCP resource to local kustomization
+  ansible.builtin.lineinfile:
+    path: "{{ audit_kustomize_dir.path }}/kustomization.yaml"
+    line: "resources:\n- oscp.yaml"
+
+- name: Apply audit logging kustomization locally
+  ansible.builtin.command: >
+    oc apply --server-side --force-conflicts -k {{ audit_kustomize_dir.path }}
+
+- name: Clean up temp directory
+  ansible.builtin.file:
+    path: "{{ audit_kustomize_dir.path }}"
+    state: absent
--- a/ci/enable-audit-logging/tasks/main.yml
+++ b/ci/enable-audit-logging/tasks/main.yml
@@ -1,0 +1,21 @@
+- name: Create service audit config secrets
+  ansible.builtin.import_tasks: create-secrets.yml
+
+# Barbican: disabled - the barbican-operator does not currently support
+# providing a custom api-paste.ini through defaultConfigOverwrite.
+# Config files are included in files/ for future use.
+# Tracked in: https://issues.redhat.com/browse/OSPRH-25781
+
+# Nova: disabled - there is a known issue with the nova-operator that
+# requires naming the audit map file as policy.yaml. Config files are
+# included in files/ for future use.
+
+- name: Copy controlplane kustomization for audit logging
+  ansible.builtin.template:
+    src: 90-kustomize-controlplane-audit-logging.yaml.j2
+    dest: "{{ cifmw_basedir }}/artifacts/manifests/kustomizations/controlplane/90-kustomize-controlplane-audit-logging.yaml"
+  when: not enable_audit_logging_local_apply
+
+- name: Apply kustomization against live OSCP
+  ansible.builtin.import_tasks: apply-local.yml
+  when: enable_audit_logging_local_apply
--- a/README.md
+++ b/README.md
@@ -59,6 +59,14 @@ This operator deploys a multiple telemetry agents, both in the control plane and
     ansible-playbook ci/deploy-logging-dependencies.yml --tags clo
     ```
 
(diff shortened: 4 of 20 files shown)


ci/enable-audit-logging/tasks/apply-local.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- name: Render kustomization to temp directory
+  ansible.builtin.template:
+    src: 90-kustomize-controlplane-audit-logging.yaml.j2
+    dest: "{{ audit_kustomize_dir.path }}/kustomization.yaml"
+
+- name: Add OSCP resource to local kustomization
+  ansible.builtin.lineinfile:
+    path: "{{ audit_kustomize_dir.path }}/kustomization.yaml"
+    line: "resources:\n- oscp.yaml"
```
vyzigold, line 20: This is a little strange looking task. We need to add
```
resources:
- oscp.yaml
```
to the file. So 2 lines of text. I was playing with different ideas. blockinfile would need to use `marker: ""` I believe. Appending it line by line with lineinfile would require 2 tasks or a loop, which seem quite unnecessary. I could also add the contents to the template guarded by the `enable_audit_logging_local_apply` variable, but I didn't want to pollute the template with it.

So I arrived at using lineinfile with \n to split the 2 lines as something I disliked the least. I can switch it to something else if there are preferences from the reviewers.

ci/enable-audit-logging/tasks/apply-local.yml
vyzigold, line 1: The role is written in a way to generate kustomization files and put them into the correct folder, where ci-framework finds them and uses them to generate the initial OpenStackControlPlane CR, which is then applied. This means audit logs will be enabled right from the start on the controlplane in the CI.

But the situation is a bit different for development and using the role / playbook locally. The classic usecase for local usage would be "We have RHOSO already deployed, let's additionally enable the audit logs".

Because of that difference between CI and local usage, we have a few tasks here to make the role work after controlplane is already deployed.

ci/enable-audit-logging/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+enable_audit_logging_namespace: "{{ cifmw_openstack_namespace | default('openstack') }}"
```
vyzigold, line 1: I decided to have a default var for the namespace here, so that I can default it to "openstack" when outside of the CI, so that we don't need to specify -e cifmw_openstack_namespace=openstack everytime we need to run this locally.

vyzigold:
Nova and Barbican will be added at later date either directly to this PR or as followup PRs based on when support in barbican-operator and nova-operator is available and when this gets through the reviews.

vyzigold:
recheck

vyzigold:
recheck no logs

vyzigold:
We can see the audit logs working in the service pod logs. Search for `http://schemas.dmtf.org/cloud/audit/1.0/event` in:
[cinder](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//943/rdoproject.org/943e59e91302440782728e66a29ecd83/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-70ce4ab587da14dddf23f3816879343650007e91038f935686ab5355e6f2a7a2/namespaces/openstack/pods/cinder-api-0/logs/cinder-api-log.log)
[glance-external](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//943/rdoproject.org/943e59e91302440782728e66a29ecd83/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-70ce4ab587da14dddf23f3816879343650007e91038f935686ab5355e6f2a7a2/namespaces/openstack/pods/glance-default-external-api-0/logs/glance-log.log)
[glance-internal](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//943/rdoproject.org/943e59e91302440782728e66a29ecd83/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-70ce4ab587da14dddf23f3816879343650007e91038f935686ab5355e6f2a7a2/namespaces/openstack/pods/glance-default-internal-api-0/logs/glance-log.log)
[keystone](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//943/rdoproject.org/943e59e91302440782728e66a29ecd83/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-70ce4ab587da14dddf23f3816879343650007e91038f935686ab5355e6f2a7a2/namespaces/openstack/pods/keystone-577f55fcff-b7wrp/logs/keystone-api.log)
[neutron](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//943/rdoproject.org/943e59e91302440782728e66a29ecd83/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-70ce4ab587da14dddf23f3816879343650007e91038f935686ab5355e6f2a7a2/namespaces/openstack/pods/neutron-66cf847db8-5f4k7/logs/neutron-api.log)

vyzigold:
recheck periodic job fix merged in https://github.com/infrawatch/feature-verification-tests/pull/379

vyzigold:
recheck this issue should be fixed again

ci/enable-audit-logging/tasks/apply-local.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- name: Apply audit logging kustomization locally
+  ansible.builtin.command: >
+    oc apply --server-side --force-conflicts -k {{ audit_kustomize_dir.path }}
+
+- name: Clean up temp directory
```
danpawlik, line 26: nit: if earler tasks would fail, it woult not remove temp file. Maybe it is not somthing "big" but worth to consider to add block + always. FOr example: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_helpers/tasks/include_file.yml#L20

ci/enable-audit-logging/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+# included in files/ for future use.
+
+- name: Copy controlplane kustomization for audit logging
+  ansible.builtin.template:
+    src: 90-kustomize-controlplane-audit-logging.yaml.j2
+    dest: "{{ cifmw_basedir }}/artifacts/manifests/kustomizations/controlplane/90-kustomize-controlplane-audit-logging.yaml"
+  when: not enable_audit_logging_local_apply
+
+- name: Apply kustomization against live OSCP
+  ansible.builtin.import_tasks: apply-local.yml
```
danpawlik, line 20: nit: include_tasks would not be better here?

danpawlik (approved):

ci/enable-audit-logging/tasks/main.yml
vyzigold: I think it's the same in this case, but why not. See followup PR

ci/enable-audit-logging/tasks/apply-local.yml
vyzigold: you're right.

PR #883, opened 2026-04-30
Simplify ClusterLogForwarder condition check

vyzigold:
Instead of counting specific numbers of True/False conditions, check that all conditions are True. This is more robust if the number of conditions changes in the future.


vyzigold:
I've seen CRs to add new conditions or remove some as their operators update. This seems to have happened with CLO recently.

elfiesmelfie (approved):

vyzigold:
recheck

PR #888, opened 2026-05-07
Enable identity_name_discovery in ceilometer compute polling config

jlarriba:
This enables adding project_name and user_name in addition to the ids as Prometheus metrics labels.

The default was set to False to avoid overwhelming keystone, but cache_utils.py already provides a caching layer (in-memory dict or redis/dogpile backend) with a 600s TTL. With caching in place, the original concern is largely mitigated, so I believe that we can have it enabled by default in RHOSO.


jlarriba:
recheck

jlarriba:
recheck

vyzigold (approved):

PR #893, opened 2026-05-13
Implement enable audit logging suggestions

vyzigold:
- explicitly specify glance backend config in customServiceConfig, because otherwise it'll get overwritten by the audit logging config
- use include_tasks instead of import tasks
- use block/always to ensure temporary dir gets always deleted

Closes: [OSPRH-30106](https://redhat.atlassian.net/browse/OSPRH-30106)


lnatapov (approved):
lgtm

PR #899, opened 2026-05-19
[OSPRH-30223] Update audit logging filter regex

vyzigold:
Change the filter regex for audit logs to the CADF schema url (http://schemas.dmtf.org/cloud/audit/1.0/event). This will prevent a potential clash in the unlikely scenario of having other notifications routed to logs. Also it allows us to use the pycadf's `use_oslo_messaging = false` option to skip oslo messaging and route audit logs into logs directly.


vyzigold:
this also aligns with the updated audit logging draft doc from last week.

jlarriba (approved):

PR #901, opened 2026-05-26
Split appcred finalizer management

Deydra71:
Following the [discussion in watcher-operator](https://github.com/openstack-k8s-operators/watcher-operator/pull/369#discussion_r3234900678)  the AC finalizer management is now split into two phases: 
* Early phase: adds consumer finalizer to the new AC secret immediately (protects it from premature revocation) 
* Late phase: removes consumer finalizer from the old AC secret only after `AllSubConditionIsTrue()`

This prevents a race condition where rapid AC rotations could revoke credentials still in use by running pods.  The same pattern is implemented in all service operators with application credential support.

### DIFF
--- a/internal/controller/autoscaling_controller.go
+++ b/internal/controller/autoscaling_controller.go
@@ -335,10 +335,13 @@ func (r *AutoscalingReconciler) reconcileNormal(
 		return rbacResult, nil
 	}
 
-	if instance.Spec.Aodh.Auth.ApplicationCredentialSecret != "" || instance.Status.ApplicationCredentialSecret != "" {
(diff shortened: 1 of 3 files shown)


Deydra71:
recheck

Looks like dataplane deployment timed out

Deydra71:
@vyzigold Hi! Is it possible to review this PR and merge so we have unified app cred finalizer related  logic in all service operators

vyzigold:
> @vyzigold Hi! Is it possible to review this PR and merge so we have unified app cred finalizer related logic in all service operators

Ah sorry! I noticed this PR and then I forgot. I'll do it before lunch.

vyzigold (approved):

PR #904, opened 2026-06-03
Don't use newest images for some OpenStack services in master CI jobs

danpawlik:
The telemetry-operator-multinode-cloudkitty CI job does not need to recreate all container images to have latest code. The most important for us it to verify claudkitty service with other telemetry related services.
That change would help to avoid unexpected issues for non-core OpenStack services during the deployment and verification.

Add all-services telemetry master CI jobs

Add two new CI jobs that verify telemetry services with all OpenStack
services enabled:

    - telemetry-openstack-meta-content-provider-all-services-master
    - telemetry-operator-multinode-master

These jobs use locally-built telemetry images instead of current-podified
tags. Since images cannot be updated before the edpm_prepare role's
"Wait for OpenStack controlplane to be deployed" task, a post-deployment
hook applies use-mcp-telemetry-containers.yml to replace current-podified
images with the local registry images built by the MCP job.

### DIFF
--- a/ci/patch-mcp-telemetry-containers.yaml.j2
+++ b/ci/patch-mcp-telemetry-containers.yaml.j2
@@ -1,0 +1,18 @@
+---
+# Patch only telemetry-related images from the MCP local registry.
+# Core OpenStack services keep install_yamls default images (current-podified).
+spec:
+  customContainerImages:
+    aodhAPIImage: {{ telemetry_registry }}/openstack-aodh-api:{{ telemetry_tag }}
+    aodhEvaluatorImage: {{ telemetry_registry }}/openstack-aodh-evaluator:{{ telemetry_tag }}
+    aodhListenerImage: {{ telemetry_registry }}/openstack-aodh-listener:{{ telemetry_tag }}
+    aodhNotifierImage: {{ telemetry_registry }}/openstack-aodh-notifier:{{ telemetry_tag }}
+    ceilometerCentralImage: {{ telemetry_registry }}/openstack-ceilometer-central:{{ telemetry_tag }}
+    ceilometerComputeImage: {{ telemetry_registry }}/openstack-ceilometer-compute:{{ telemetry_tag }}
+    ceilometerIpmiImage: {{ telemetry_registry }}/openstack-ceilometer-ipmi:{{ telemetry_tag }}
+    ceilometerNotificationImage: {{ telemetry_registry }}/openstack-ceilometer-notification:{{ telemetry_tag }}
+    cloudkittyAPIImage: {{ telemetry_registry }}/openstack-cloudkitty-api:{{ telemetry_tag }}
+    cloudkittyProcImage: {{ telemetry_registry }}/openstack-cloudkitty-processor:{{ telemetry_tag }}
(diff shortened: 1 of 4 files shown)


vyzigold:
I think, in the end, we should have 2 content provider jobs. This one, which would build just the services that we need updated, because we need to test newer features together with the rest of RHOSO. That'd be aodh, ceilometer, cloudkitty, heat and I think tempest-all to have the new autoscaling tests. Basically the containers listed in https://github.com/openstack-k8s-operators/telemetry-operator/blob/f9d46da27a8b12e7af63fd361b12e4f16ebf1457/ci/files/containers.yaml . We want to leave the rest of the images to be antelope, since that's the closest to RHOSO 18.

Then in future PR, we should add a second content provider job, which would build all the services from master to have early testing for RHOSO 19.

This would mean, that IMO you're missing some additional images to exclude, like keystone, nova, neutron, cinder, glance (maybe I missed some additional ones)

danpawlik:
recheck

danpawlik:
let's discuss that after the weekend :)

danpawlik:
@vyzigold hey, updated. Hope I understand correctly your suggestion

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
workflow is more less:
```
edpm_prepare : Wait for OpenStack controlplane to be deployed
```

then

```
2026-06-11 10:23:37.540269 | controller | TASK [run_hook : Loop on hooks for post_ctlplane_deploy _raw_params={{ hook.type }}.yml] ***
2026-06-11 10:23:37.540278 | controller | Thursday 11 June 2026  10:23:37 +0000 (0:00:00.173)       0:15:39.877 *********
2026-06-11 10:23:37.540282 | controller | Thursday 11 June 2026  10:23:37 +0000 (0:00:00.173)       0:15:39.876 *********
2026-06-11 10:23:37.782987 | controller | included: /home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/run_hook/tasks/playbook.yml for localhost => (item={'source': '/home/zuul/src/github.com/openstack-k8s-operators/telemetry-operator/ci/use-mcp-telemetry-containers.yml', 'type': 'playbook', 'name': '99 patch mcp telemetry containers'})
2026-06-11 10:23:37.783022 | controller | included: /home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/run_hook/tasks/playbook.yml for localhost => (item={'name': 'Tune rabbitmq resources', 'type': 'playbook', 'source': 'rabbitmq_tuning.yml'})
```

In other words, first it do in stable `current-podified` then it update to tcib image `telemetry-openstack-meta-content-provider-master`

danpawlik:
alternative way is to build all container images and just disable operators that are not used, but still we will be using "fresh" images.

danpawlik:
@sdatko suggest to go with similar approach as Watcher is doing https://github.com/openstack-k8s-operators/watcher-operator/blob/main/.zuul.yaml#L155

danpawlik:
It is replaced by https://github.com/openstack-k8s-operators/telemetry-operator/pull/915.

PR #905, opened 2026-06-05
dashboard: add SR-IOV VF metrics to network exporter dashboard

aharivel:
Add a new SR-IOV row to the OpenStack Network Exporter dashboard with panels for VF configuration, traffic bytes/packets rates, dropped packets, and broadcast/multicast counters.

The related commits **already merged** that enable networking SR-IOV metrics in the openstack network exporter are the following:
- https://github.com/openstack-k8s-operators/openstack-network-exporter/commit/135ad54161df54044ed42d0ef43cc2b018aa444a
- https://github.com/openstack-k8s-operators/openstack-network-exporter/commit/474933e4b46062f080ebb7a5b408f3310d0c3d84

### DIFF
--- a/internal/dashboards/openstack-openstack-network.go
+++ b/internal/dashboards/openstack-openstack-network.go
@@ -22,7 +22,7 @@ import (
 )
 
 // OpenstackOpenstackNetwork creates a ConfigMap containing the OpenStack network monitoring dashboard
-func OpenstackOpenstackNetwork(dsName string, hasDpdk bool) *corev1.ConfigMap {
+func OpenstackOpenstackNetwork(dsName string, hasDpdk bool, hasSriov bool) *corev1.ConfigMap {
 	dashBoardOVN :=
 		`{
         "__inputs": [],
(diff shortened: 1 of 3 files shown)


internal/dashboards/openstack-openstack-network.go
```diff
@@ -1133,6 +1626,7 @@ func OpenstackOpenstackNetwork(dsName string, hasDpdk bool) *corev1.ConfigMap {
 	if hasDpdk {
 		dashBoard += "," + dashBoardDpdk
 	}
+	dashBoard += "," + dashBoardSRIOV
```
vcandapp: @aharivel, should we add if condition "has Sriov" like dpdk above ?

aharivel: Thanks Viji, I made some modification. 
SR-IOV VF panels is now only showed when neutron-sriov service is enabled on at least one nodeset. 
I followed the dpdk pattern.

internal/dashboards/openstack-openstack-network.go
```diff
@@ -1059,6 +1059,499 @@ func OpenstackOpenstackNetwork(dsName string, hasDpdk bool) *corev1.ConfigMap {
...
+                        "pointradius": 2,
+                        "points": false,
+                        "renderer": "flot",
+                        "seriesOverrides": [],
+                        "spaceLength": 10,
+                        "stack": false,
+                        "steppedLine": false,
+                        "targets": [
+                            {
+                                "expr": "rate(net_vf_receive_packets_total{instance=\"$instance\"}[$__rate_interval])",
```
vcandapp, line 1308: Of the 5 panels proposed, only the "VF configuration" might be valid even when VFs are not allocated. Just enquiring about the other panels like "VF traffic" , "VF packets" .... etc - will they be visible for "unallocated-VF"s , just showing zeros ? Or the PromqL query "rate(net_vf_transmit_packets_total{instance=\"$instance\"}[$__rate_interval])", will handle this ?

aharivel: Good question. All VFs emit net_vf_*_total metrics regardless of allocation status; I check this on my lab test. For unallocated VFs (mac 00:00:00:00:00:00, link_state auto), the counters stay at 0, so rate() returns 0 and the panels show flat zero lines.
  
This is intentional, it gives operators visibility into which VFs are active vs idle. The "VF Configuration" table already shows allocation details (MAC, VLAN, link_state), so the zero-traffic VFs are easy to correlate.

If we want to filter out unallocated VFs from traffic panels, we could join with net_vf_info{link_state="enable"} in the PromQL queries, but I think showing all VFs is more useful for monitoring, you'd notice unexpected traffic on supposedly unallocated VFs, which could indicate a misconfiguration.

Let me know if you'd prefer filtering them out.

vcandapp: @aharivel , few follow-on questions: 

- **we could join with net_vf_info{link_state="enable"}**: more useful from user perspective would be able to differentiate between "link_state enabled + idle" and "link_state enabled + down" - ie can we clearly distinguish between idle vs down VFs based on interpration of counters ?
- for you point on **"you'd notice unexpected traffic on supposedly unallocated VFs"** - do the counters persist on the VF device on instance create/delete ?
- Am fine if we can create a sperate Jira for such enhancements and do in a follow-up PR

aharivel: Agin good questions !

1. **Idle vs down**: `link_state` in `net_vf_info` reflects the admin state (`auto`/`enable`/`disable`), not the operational link state. So with counters alone we can't distinguish "enabled + idle" from "enabled + link down". We'd need the kernel's operational state which isn't currently exposed by rtnetlink VF info. This could be a follow-up enhancement; adding an operational state label to `net_vf_info` in the exporter itself.

2. **Counter persistence**: Yes, VF counters are maintained by the PF driver in the kernel and persist across VM create/delete. They only reset on PF driver reload or host reboot. So `rate()` handles this correctly - it shows the current traffic rate regardless of previous instances.

3. Agreed - I'll create a Jira for the filtering/state enhancements as a follow-up.

jlarriba (approved):

internal/dashboards/openstack-openstack-network.go
```diff
@@ -1059,6 +1059,499 @@ func OpenstackOpenstackNetwork(dsName string, hasDpdk bool) *corev1.ConfigMap {
...
+                "collapse": false,
+                "collapsed": false,
+                "panels":
+                [
+                    {
+                      "datasource": {
+                          "name": "` + dsName + `",
+                          "type": "prometheus"
+                      },
+                      "description": "SR-IOV Virtual Function configuration",
```
vcandapp, line 1073: @aharivel - another relevant question on this table: What about NIC-partitioned case where some VFs are used by host ? Is it worth to distinguish them here ?

aharivel: @vcandapp In a second phase, we will correlate with Neutron/Nova which VF is used by OSP and for which purposes. That means the left over will be the one used by the Host for which ever purposes other than OSP.

vcandapp: Sure, thanks @aharivel

vcandapp (approved):
The updates lgtm

PR #913, opened 2026-06-15
[OSPRH-31346] CI - Use audit map files from /etc/pycadf for enabling audit logging

vyzigold:
Turns out the audit map files are already available inside the pods inside /etc/pycadf. Use those instead of providing them externally when enabling the audit logging in CI. This will align a little better with the WIP documentation.

Generated-By: Claude Opus 4.6


vyzigold:
recheck

danpawlik:
suggest to improve commit message

vyzigold:
> suggest to improve commit message

better?

danpawlik:
recheck

vyzigold:
recheck looks like 502 from github api

SeanMooney (approved):

SeanMooney:
reviewing this in more detail i dont think this will work for nova
it no less wrong then before
but you ahve not added the nova_api_audit_map.conf 
via defaultConfigOverwrite

SeanMooney (commented):
so ya https://github.com/openstack-k8s-operators/nova-operator/pull/1104/changes#diff-d8540b18201f7d68bc4c3e4beef4c4911466573e5542799e3d85844bfacc86b6R108 you need to pass it inlien for nova

and 

https://github.com/openstack-k8s-operators/telemetry-operator/pull/914/changes#diff-dea4161a1cf9d3334e35c139b7a3f8a09955713fcfc91cb7005c74513339ee3b 

dose not do that

also the destiantionis not configurabel in nova

https://github.com/openstack-k8s-operators/nova-operator/pull/1104/changes#diff-03742c5c10c165b2d6435a64b73e30b58c2c0890a8f726b14d279dc832c55d05R89

https://github.com/openstack-k8s-operators/nova-operator/blob/main/templates/nova/api/config/nova-api-config.json#L89

so this is incorrect or at least its incomplete

ci/enable-audit-logging/files/nova-api-paste.ini
```diff
@@ -91,6 +91,6 @@ paste.filter_factory = keystonemiddleware.auth_token:filter_factory
 
 [filter:audit]
 paste.filter_factory = keystonemiddleware.audit:filter_factory
-audit_map_file = /etc/nova/nova_api_audit_map.conf
+audit_map_file = /etc/pycadf/nova_api_audit_map.conf
```
SeanMooney, line 94: just so its clear this wont work for nova. since the operator wont put the file there
and #914 does not pass the file to the api inlen and nova does not supprot extra mounts by design. so you cant circumvent it that way.

vyzigold: I understand what you mean. But last Wednesday we discovered one thing (unfortunately it took us until last Wednesday to realize it). pycadf is a requirement of keystonemiddleware, so anything that has keystonemiddleware installed also has pycadf installed, which means the audit_map files are put in /etc/pycadf/ in all the api containers we care about. This way it actually makes documentation a lot simpler, instead of documenting where to get the file and how to pass it into pods (through the defaultConfigOverwrite or extraMounts), it's already there and we just need to point the api-paste at the correct path.

You can see this working in the nova pod inside the telemetry-operator-multinode-audit-logging job, which is using the changes from this PR and it should also be the same ansible that's used in the watcher real data job. https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//f97/rdoproject.org/f97e77fb288c4ad18294a97e1fd58b4d/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-7459e6a61054bff131dec69759ac45c73c773a5e3113920237ed8b77cb4fa4f8/namespaces/openstack/pods/nova-api-0/logs/nova-api-log.log (you can even see audit logs appearing if you find for `http://schemas.dmtf.org/cloud/audit/1.0/event`)

PR #915, opened 2026-06-16
Don't use newest images for some OpenStack services in master CI jobs

danpawlik:
The telemetry-operator-multinode-cloudkitty CI job does not need to recreate all container images to have latest code. The most important for us it to verify claudkitty service with other telemetry related services.

Also add two new CI jobs that verify telemetry services with all OpenStack services enabled:

    - telemetry-openstack-meta-content-provider-all-services-master
    - telemetry-operator-multinode-master

These jobs use locally-built telemetry images instead of current-podified tags. Since images cannot be updated before the edpm_prepare role's "Wait for OpenStack controlplane to be deployed" task, a post-deployment hook applies use-mcp-telemetry-containers.yml to replace current-podified images with the local registry images built by the MCP job.

Assisted-by: Clause 4.6

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -1,14 +1,11 @@
 ---
 - job:
-    name: telemetry-openstack-meta-content-provider-master
+    name: telemetry-openstack-meta-content-provider-all-services-master
     parent: openstack-meta-content-provider-master
     description: |
-      A meta content provider zuul job to build telemetry
-      specific containers.
+      Meta content provider that builds telemetry-operator images and all
+      OpenStack containers from master for early RHOSO 19 testing.
     vars:
-      # Note(Chandan Kumar): image_base is the operator name without -operator.
-      # It is needed by openstack-operator to include telemetry operator images
-      # in the openstack-operator catalog image.
       cifmw_build_containers_force: true
       cifmw_bop_dlrn_from_source: true
       cifmw_build_containers_registry_namespace: podified-master-centos10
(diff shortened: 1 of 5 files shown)


danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

sdatko (approved):
/lgtm

zuul.d/projects.yaml
```diff
@@ -142,17 +166,40 @@
...
       fetch_dlrn_hash: false
+      # Skip full update_containers; patch only telemetry images
+      # from meta content provider after control plane deploy. Core services keep stable images.
+      cifmw_prepare_openstackversion: false
+      cifmw_extras:
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir }}/scenarios/centos-9/multinode-ci.yml"
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-cloudkitty-tempest.yml"
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-use-meta-content-telemetry-containers.yml"
+    roles:
+      - zuul: github.com/openstack-k8s-operators/ci-framework
```
sdatko, line 189: This may be not necessary ~ i think (to be verified) one of the parents already defines those roles, but it is fine to keep it

vyzigold: I agree, I'd expect this to already be in some of the parent.

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
recheck Seems like the FVT autoscaling scenario, which is very timing sensitive failed to execute correctly. It's loading 1 VM and waiting until it scales to 3 VMs. But it seems like the cpu_low alarm triggered immaturely, which removed the leaded VM and then the stack stayed with just 1 VM without any load, thus the success condition was never reached.

vyzigold (approved):

PR #922, opened 2026-06-23
Revert identity name discovery

vyzigold:
We discovered, that ceilometer version we use has a bug, where it tries to reach Keystone through the public endpoint no matter what's the configuration when performing the identity name discovery. We can't assume the public service endpoints are reachable from the compute nodes. So having `identity_name_discovery=True` started causing issues in environments without access to public endpoint from the computes. This caused ceilometer compute agent to completely fail and not produce any metrics. We'll need to revert these 2 commits until ceilometer is fixed to adhere to its configuration, so that we can unblock further development of other features.


vyzigold:
recheck failed image pull

elfiesmelfie (approved):

PR #923, opened 2026-06-24
[18.0-fr6] Revert identity name discovery

vyzigold:
We discovered, that ceilometer version we use has a bug, where it tries to reach Keystone through the public endpoint no matter what's the configuration when performing the identity name discovery. We can't assume the public service endpoints are reachable from the compute nodes. So having identity_name_discovery=True started causing issues in environments without access to public endpoint from the computes. This caused ceilometer compute agent to completely fail and not produce any metrics. We'll need to revert these 2 commits until ceilometer is fixed to adhere to its configuration, so that we can unblock further development of other features.

cherry-pick of: https://github.com/openstack-k8s-operators/telemetry-operator/pull/922

Depends-On: https://github.com/openstack-k8s-operators/watcher-operator/pull/390
Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/924

### DIFF
--- a/internal/dashboards/openstack-network-traffic.go
+++ b/internal/dashboards/openstack-network-traffic.go
@@ -109,12 +109,12 @@ func OpenstackNetworkTraffic(dsName string) *corev1.ConfigMap {
                                 "steppedLine": false,
                                 "targets": [
                                     {
-                                        "expr": "vm:ceilometer_network_incoming_bytes:rate1m{project_name =~ \"$project\", vm_name =~ \"$VM\"}",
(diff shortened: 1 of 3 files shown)


vyzigold:
recheck

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
I've checked the zuul logs. All jobs are green except for functional-tests-osp18, which is running must-gather. From reading the logs, tempest tests passed, I also don't see a failed task from FVT, so we should be good to merge.

Regarding the 2 Depends-On, it's just to make the Zuul work on the new FR6 branch, no actual changes to operator code, so we can merge this PR even without the Depends-On being merged.

elfiesmelfie (approved):

PR #925, opened 2026-06-25
Set service_type in [keystone_authtoken] for access rule validation

Deydra71:
Without service_type configured, keystonemiddleware cannot validate application credentials with custom access rules, causing HTTP 401 for end users.

Closes: [OSPRH-22365](https://redhat.atlassian.net/browse/OSPRH-22365)

### DIFF
--- a/templates/autoscaling/config/aodh.conf
+++ b/templates/autoscaling/config/aodh.conf
@@ -60,6 +60,7 @@ project_domain_name = Default
 {{- end }}
 auth_url = {{ .KeystoneInternalURL }}
 service_token_roles_required = True
+service_type = alarming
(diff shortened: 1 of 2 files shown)


vyzigold:
We don't need similar for the service_credentials config sections?

Deydra71:
@vyzigold No, because this is only for incoming requests to the API, it doesn't affect outgoing requests to other services

https://docs.openstack.org/keystone/latest/user/application_credentials.html#access-rules

vyzigold:
The change for Aodh probably isn't needed, since Keystone will never support App creds with trust. But this is fine as is and I'll figure out Aodh later.

vyzigold (approved):

PR #936, opened 2026-07-09
add ScrapeConfig for OpenStack Lightspeed

jancervenka:
add OpenStack Lightspeed ScrapeConfig using a modified createServiceScrapeConfigFromLabelSelector function that accepts a specific targetNamespace where to look for the service


internal/controller/metricstorage_controller.go
```diff
@@ -1100,8 +1104,118 @@ func (r *MetricStorageReconciler) createOVSDBServerNBScrapeConfig(
...
+	})
+	if err != nil {
+		return err
+	}
+	if op != controllerutil.OperationResultNone {
+		Log.Info(fmt.Sprintf("OpenStack Lightspeed metrics token Secret %s successfully changed - operation: %s", tokenSecret.Name, string(op)))
+	}
+
+	crbName := fmt.Sprintf("%s-%s", metricstorage.OpenStackLightspeedAccessCRName, instance.Namespace)
+	crb := &rbacv1.ClusterRoleBinding{
```
jlarriba, line 1292: This ClusterRoleBinding cannot have a ControllerReference being a cluster resource so it will not be cleaned up automatically when the CR is removed.

You should add explicit delete code for this CRB in the reconcileDelete function so it does not linger there forever.

internal/controller/metricstorage_controller.go
```diff
@@ -1100,8 +1104,118 @@ func (r *MetricStorageReconciler) createOVSDBServerNBScrapeConfig(
...
+		return controllerutil.SetControllerReference(instance, caConfigMap, r.Scheme)
+	})
+	if err != nil {
+		return err
+	}
+	if op != controllerutil.OperationResultNone {
+		Log.Info(fmt.Sprintf("OpenStack Lightspeed CA bundle ConfigMap %s successfully changed - operation: %s", caConfigMap.Name, string(op)))
+	}
+
+	sa := &corev1.ServiceAccount{
```
jlarriba, line 1254: All these resources (ConfigMap, ServiceAccount, Secret, and ClusterRoleBinding) are created on every reconcile, even then openstack-lightspeed does not exist on the cluster.

It would be better to do the same that you do with the ScrapeConfig and condition their creation to openstack-lightspeed existence.

internal/controller/metricstorage_controller.go
```diff
@@ -1100,8 +1104,118 @@ func (r *MetricStorageReconciler) createOVSDBServerNBScrapeConfig(
...
+		Log.Info(fmt.Sprintf("OpenStack Lightspeed metrics ClusterRoleBinding %s successfully changed - operation: %s", crb.Name, string(op)))
+	}
+
+	labelSelector := map[string]string{
+		"metrics": "enabled",
+		"service": "lightspeed-app-server",
+	}
+	openStackLightspeedCfgName := fmt.Sprintf("%s-openstack-lightspeed", telemetry.ServiceName)
+	return r.createServiceScrapeConfigFromLabelSelector(
+		ctx, instance, helper, serviceLabels, "openstack-lightspeed",
```
jlarriba: Hardcoding the namespace does not seem wise, any reason why this could not be a MetricStorage spec variable with an appropiate default value, but that the user can change if he wants to?

jancervenka:
@jlarriba thank you for the review Juan! I implemented the suggestions, can you please take a look when you have the time?

internal/controller/metricstorage_controller.go
jancervenka: I added a variable `OpenStackLightspeedNamespace` with the default value `"openstack-lightspeed"` to the MetricStorage spec.

internal/controller/metricstorage_controller.go
jancervenka: I added this logic to the `createOpenStackLightspeedScrapeConfig` function that returns early before any resources are created if the lightspeed service is not found:

```golang
	err := helper.GetClient().List(ctx, serviceList, listOpts...)
	if err != nil {
		Log.Info(fmt.Sprintf("Cannot get OpenStack Lightspeed metrics services. Scrape configs not created. Error: %s", err))
		return nil
	}
	if len(serviceList.Items) == 0 {
		Log.Info("No OpenStack Lightspeed metrics services found")
		return nil
	}
```

internal/controller/metricstorage_controller.go
jancervenka: I added the following logic to the `reconcileDelete` function that deletes the CRB.
```golang
	crbName := fmt.Sprintf("%s-%s", metricstorage.OpenStackLightspeedAccessCRName, instance.Namespace)
	crb := &rbacv1.ClusterRoleBinding{}
	err := r.Get(ctx, types.NamespacedName{Name: crbName}, crb)
	if err != nil {
		if !k8s_errors.IsNotFound(err) {
			return ctrl.Result{}, err
		}
	} else {
		if err := r.Delete(ctx, crb); err != nil && !k8s_errors.IsNotFound(err) {
			return ctrl.Result{}, err
		}
		Log.Info(fmt.Sprintf("Deleted OpenStack Lightspeed ClusterRoleBinding %s", crbName))
	}
```

jancervenka:
recheck

jancervenka:
recheck

jancervenka:
recheck

jlarriba (approved):

PR #941, opened 2026-07-13
Disable event processing in ceilometer notification agent

jlarriba:


vyzigold:
I was interested why the default config would send events as notifications.

There is a "notifier" defined here https://opendev.org/openstack/ceilometer/src/branch/master/setup.cfg#L175
which leads to: https://opendev.org/openstack/ceilometer/src/branch/master/ceilometer/publisher/messaging.py#L239

Either way, this LGTM.

vyzigold (approved):

jlarriba:
/retest

openstack-operator-index image build timed out

PR #947, opened 2026-07-17
[OSPRH-32283] Make polling.yaml dynamic based on enabled services

vyzigold:
The ceilometer central agent's polling.yaml will change dynamically and include Designate, Octavia or Manila polling configuration if these services are enabled. The mechanism is watching the KeystoneEndpoint CRs to determine whether these services are enabled. This means we don't need to add any new watches (like watching the Octavia CR) and we don't need any new dependencies (like the octavia-operator). This works very similar to how swift polling config worked until now. Also ceilometer would already restart on each new KeystoneEndpoint anyway, so there isn't any additional reconciling / pod restarts compared to the previous behavior.

### DIFF
--- a/internal/controller/ceilometer_controller.go
+++ b/internal/controller/ceilometer_controller.go
@@ -1309,7 +1309,10 @@ func (r *CeilometerReconciler) generateServiceConfig(
 		"TransportURL":        string(transportURLSecret.Data["transport_url"]),
 		"CeilometerPassword":  string(ceilometerPasswordSecret.Data["CeilometerPassword"]),
 		"TLS":                 false, // Default to false. Change to true later if TLS enabled
-		"SwiftRole":           false, //
+		"SwiftRole":           false,
+		"ManilaEnabled":       false,
+		"DesignateEnabled":    false,
+		"OctaviaEnabled":      false,
 		"Timeout":             instance.Spec.APITimeout,
 		"Region":              keystoneAPI.GetRegion(),
 	}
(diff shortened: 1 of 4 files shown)


internal/controller/ceilometer_controller.go
```diff
@@ -1309,7 +1309,10 @@ func (r *CeilometerReconciler) generateServiceConfig(
 		"TransportURL":        string(transportURLSecret.Data["transport_url"]),
 		"CeilometerPassword":  string(ceilometerPasswordSecret.Data["CeilometerPassword"]),
 		"TLS":                 false, // Default to false. Change to true later if TLS enabled
-		"SwiftRole":           false, //
+		"SwiftRole":           false,
+		"ManilaEnabled":       false,
```
jlarriba, line 1313: Well, the idea was more to retrieve the services from Keystone and dynamically adjust the ceilometer configuration. Passing manually enabled/not enabled seems to be much more clunky IMHO

jlarriba: Na, i was misunderstanding the whole PR, sorry its friday.

vyzigold:
recheck

internal/controller/ceilometer_controller.go
```diff
@@ -2017,7 +2022,7 @@ func (r *CeilometerReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 		).
 		Watches(
 			&keystonev1.KeystoneEndpoint{},
-			handler.EnqueueRequestsFromMapFunc(keystoneEndpointsWatchFn),
+			handler.EnqueueRequestsFromMapFunc(r.findObjectForSrc),
```
vyzigold, line 2023: r.findObjectForSrc is functionally almost the same as keystoneEndpointWatchFn. The difference is, that keystoneEndpointsWatchFn reconciled only the `ceilometer.ServiceName` named ceilometer. While r.findObjectForSrc reconciles all ceilometer CRs in a given namespace, making it independent on the CR name. This is more correct and it's needed for kuttl tests to work, since the ceilometer there is called `telemetry-kuttl`.

vyzigold:
recheck

vyzigold:
recheck seems to me like github or zuul is having some issues

vyzigold:
recheck

vyzigold:
recheck Failed to download packages: tmux-3.2a-5.el9.x86_64

internal/controller/ceilometer_controller.go
```diff
@@ -2097,6 +2100,24 @@ func (r *CeilometerReconciler) findObjectForSrc(ctx context.Context, src client.
...
+	ctx context.Context,
+	h *helper.Helper,
+	endpointName string,
+	namespace string,
+	log logr.Logger,
+) bool {
+	_, err := keystonev1.GetKeystoneEndpointWithName(ctx, h, endpointName, namespace)
+	if err != nil {
+		if !k8s_errors.IsNotFound(err) {
+			log.Info("Failed to check for KeystoneEndpoint, assuming not present",
```
jlarriba: Minor, but shouldnt this be a log.Error trace?

vyzigold: I wasn't sure which level would be better. IIRC we've historically gravitated more towards the less severe error levels. But you're right that this situation kind of sounds like an error. I'll switch it to error :+1:

vyzigold:
recheck

vyzigold:
recheck

jlarriba (approved):

PR #951, opened 2026-07-20
Add vcpus and memory pollers to compute polling config

jlarriba:


danpawlik (approved):
LGTM but the commit message is too minimalistic

sdatko (approved):
/lgtm /approve

PR #969, opened 2026-07-30
[OSPRH-33382] Generate telemetry_secret per-deployment instead of har…

jlarriba:
…dcoded literal

The Ceilometer publisher telemetry_secret was hardcoded as a static literal in three config templates (central, compute, ipmi), meaning every deployment shared the same publicly known HMAC key. This replaces it with a per-deployment random value generated on first reconcile and persisted via CreateOrPatchSecretPreserve, following the same pattern used for other templated secrets in the operator.


internal/controller/ceilometer_controller.go
```diff
@@ -1269,11 +1281,50 @@ func (r *CeilometerReconciler) reconcileKSM(
...
+	instance *telemetryv1.Ceilometer,
+) (string, error) {
+	const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	b := make([]byte, 25)
+	if _, err := rand.Read(b); err != nil {
+		return "", fmt.Errorf("error generating telemetry secret: %w", err)
+	}
+	for i := range b {
+		b[i] = charset[int(b[i])%len(charset)]
+	}
```
vyzigold: I tried to look through ceilometer code if the telemetry_secret has any charset or length restrictions and I don't see any. As far as I see, in the end it just ends up being a string, which gets encoded in utf-8 and that's fed to hmac.new: https://github.com/openstack/ceilometer/blob/d550704bbc8b425d17d77db2689ad4a494221b48/ceilometer/publisher/utils.py#L70 (other than this function, it seems to be just passed through the code from function to function)

So I wondered if this whole algorithm could be simplified and if we could instead use some built in function provided by one of our libraries here. What about this claude suggestion:


```suggestion
	b := make([]byte, 25)
	if _, err := rand.Read(b); err != nil {
    	return "", fmt.Errorf("error generating telemetry secret: %w", err)
	}
	secret := hex.EncodeToString(b)
```
This has no charset const and no iteration through `b` + looping through `charset` array. Then you'd use the `secret` string below.

jlarriba: good advice, thank you!

elfiesmelfie:
recheck

danpawlik (approved):

PR #971, opened 2026-07-30
[OSPRH-33383] Remove unused privileged SCC from CloudKitty kubebuilde…

jlarriba:
…r RBAC marker

The CloudKitty controller's kubebuilder marker requested use on both anyuid and privileged SCCs, but no generated workload sets Privileged: true or uses host namespaces. The workload Role the controller creates only delegates anyuid. This removes the dead-weight privileged grant from the marker and regenerates config/rbac/role.yaml.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -322,15 +322,6 @@ rules:
   verbs:
   - use
 - apiGroups:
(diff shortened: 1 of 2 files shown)


jlarriba:
recheck

PR #975, opened 2026-08-03
Include ssl.conf in httpd configuration

fmount:
The `ssl.conf` rendered by `lib-common` was placed by `kolla` into `conf.d/` but never loaded because the Include `conf.d/*.conf` directive was commented out. This left global `SSL` hardening settings (cipher suite, protocol restrictions, session cache) at `mod_ssl` defaults instead of the operator-managed values.

Replace the explicit Include of `00wsgi-<service>.conf` with the `conf.d` wildcard so both `ssl.conf` and the `wsgi vhost` config are loaded.

Jira: https://redhat.atlassian.net/browse/OSPRH-34160

### DIFF
--- a/templates/autoscaling/config/httpd.conf
+++ b/templates/autoscaling/config/httpd.conf
@@ -21,7 +21,4 @@ CustomLog /dev/stdout combined env=!forwarded
 CustomLog /dev/stdout proxy env=forwarded
 ErrorLog /dev/stdout
 
-# XXX: To disable SSL
-#Include conf.d/*.conf
-# If above include is commented include at least the aodh wsgi file
-Include conf.d/00wsgi-aodh.conf
+Include conf.d/*.conf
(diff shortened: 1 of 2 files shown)


stuggi:
recheck

fmount:
recheck

stuggi:
recheck

vyzigold:
This seems to fail with the same webhook issue across all jobs across rechecks. I think I've seen this failure before and a rebase helped. Can you try to rebase your branch on top of main?

fmount:
> This seems to fail with the same webhook issue across all jobs across rechecks. I think I've seen this failure before and a rebase helped. Can you try to rebase your branch on top of main?

Thanks, rebased!

vyzigold (approved):
as long as the zuul tests pass, this lgtm. The CI seems much better now after the rebase.

PR #976, opened 2026-08-04
[OSPRH-33507] Allow using mTLS for scraping metrics

vyzigold:
This adds a new PrometheusClientCertSecret field in the metricStorage section of the CRDs. When this field is set, it causes all scrapeConfigs created by telemetry-operator to provide the client certificate for each scrape (except the scrapeconfigs for rabbitmq, which I don't expect having mTLS support at this time).

The goal is to have the openstack-operator generate the client certificates and configure the PrometheusClientCertSection.


vyzigold:
I've done testing in a manually configured mixed environment, where some metric exporters had the current config (no mTLS required) and some exporters (ksm and prometheus-podman-exporter) required mTLS and they checked the client certs against the openstack CA. In my testing I saw that:

- Prometheus was able to correctly scrape exporters requiring client certs.
- Exporters that didn't require the client certs continued to be scraped without any issues. The client certs were provided in the requests, but they were silently ignored by the exporters. This means we have an easy way to configure mTLS across any exporters we want. We first merge this PR, then a PR for generating the client certs in openstack-operator and then we can configure the exporters whenever we want.

internal/controller/metricstorage_controller.go
```diff
@@ -575,6 +577,28 @@ func (r *MetricStorageReconciler) reconcileNormal(
 		}
 	}
 
+	// Validate Prometheus client cert secret for mTLS if provided
+	if instance.Spec.PrometheusClientCertSecret.Enabled() {
```
jlarriba: This code block runs when the CertSecret exists even when TLS is disabled. If a user sets prometheusClientCertSecret without enabling prometheusTls, the reconciler either blocks for no reason or reports false success

internal/metricstorage/scrape_config.go
jlarriba, line 67: The MySQL Exporter endpoint is not protected with authentication? I can't see the mTLS applied here, but also I can't seem to find any other auth.

vyzigold: The mysqld_exporter's scrape config will inherit these mTLS related changes thanks to L65. Currently the mysqld_exporter scrape endpoint isn't authenticated, mTLS can be enabled later with the same config change as prometheus-exporter or KSM.

internal/controller/metricstorage_controller.go
vyzigold: You're right. The intention was to enable the client side of mTLS only when the server side is enabled as well. I'm missing a condition here. Will fix.

jlarriba:
recheck

jlarriba (approved):

PR #978, opened 2026-08-06
Set sg-core prometheus plugin listen address to localhost

vyzigold:
The httpd proxy is the only thing connecting to the sg-core prometheus scrape endpoint. Httpd runs inside the same pod, so there is no reason for sg-core's prometheus plugin to listen on more than just the loopback network.


vyzigold:
recheck seems like errors claiming the cluster in prow and error downloading epel-low-priority metadata in zuul

PR #982, opened 2026-08-11
Add transport URL secret rotation with consumer finalizer

lmiccini:
When infra-operator rotates a RabbitMQ transport URL (creating a new secret and user), consumer operators must hold a consumer finalizer on the old secret until all their pods have rolled out with the new credentials. Without this, infra-operator cleans up the old RabbitMQ user while pods are still connected with old credentials, causing message bus outages.

Design:

1. Add consumer finalizer to the current transport URL secret early in reconcile. Set instance.Status.TransportURLSecret for first-time setup only (when empty or unchanged); during rotation the status is updated solely by FinalizeSecretRotation at end of reconcile.

2. Pass transportURL.Status.SecretName directly to sub-CR creation functions and config generation as a parameter — never read from instance.Status.TransportURLSecret for sub-CR specs.

3. Use statefulset.IsReady() / deployment.IsReady() from lib-common in all sub-CR controllers for accurate rollout status.

4. Use object.ManageRotationGracePeriod() to enforce a 60-second grace period before evaluating the rotation guard. This gives sub-CRs time to detect config changes, update their workloads, and roll pods — without relying on informer cache freshness.

5. Guard: CredentialRotationGuardReady(true, conditions) — evaluates AllSubConditionIsTrue after the grace period expires. Only when all sub-CR conditions are True does FinalizeSecretRotation remove the consumer finalizer from the old secret.

The same pattern applies to notification transport URL secrets and application credential secrets where applicable.


lmiccini:
recheck

internal/controller/autoscaling_controller.go
```diff
@@ -723,22 +762,48 @@ func (r *AutoscalingReconciler) reconcileNormal(
...
-				return ctrl.Result{}, err
-			}
-			instance.Status.ApplicationCredentialSecret = instance.Spec.Aodh.Auth.ApplicationCredentialSecret
-		}
-	} else {
-		instance.Status.ApplicationCredentialSecret = instance.Spec.Aodh.Auth.ApplicationCredentialSecret
+	rotationPending := false
+	if instance.Status.NotificationsURLSecret != nil {
+		rotationPending = *instance.Status.NotificationsURLSecret != "" &&
+			*instance.Status.NotificationsURLSecret != currentNotifSecret
```
jlarriba: The instance.Status.NotificationsURLSecret variable is unconditionally overwritten before rotationPending is calculated and it never changes so that means that here, the "*instance.Status.NotificationsURLSecret != currentNotifSecret" condition is always false, and thus this controller never activates the 60-second grace period.

This also happens in ceilometer_controller, but it is correctly handled in the cloudkitty_controller.

This might do the trick:

```suggestion
	rotationPending := oldNotifSecret != "" && oldNotifSecret != currentNotifSecret
```

internal/controller/ceilometer_controller.go
```diff
@@ -845,38 +887,72 @@ func (r *CeilometerReconciler) reconcileCeilometer(
...
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DeploymentReadyRunningMessage))
+	}
+
+	rotationPending := false
+	if instance.Status.NotificationsURLSecret != nil {
+		rotationPending = *instance.Status.NotificationsURLSecret != "" &&
+			*instance.Status.NotificationsURLSecret != currentNotifSecret
+	}
```
jlarriba: Same as in autoscaling, this could be:


```suggestion
    rotationPending := oldNotifSecret != "" && oldNotifSecret != currentNotifSecret
```

lmiccini:
recheck

api/v1beta1/telemetry_consts.go
```diff
@@ -18,6 +18,11 @@ package v1beta1
 import "time"
 
 const (
+	// TelemetryTransportConsumerFinalizer is the consumer finalizer added to
+	// transport secrets by all telemetry controllers (CloudKitty, Autoscaling,
+	// Ceilometer) to prevent premature deletion during credential rotation.
+	TelemetryTransportConsumerFinalizer = "openstack.org/telemetry-transport-consumer"
```
jlarriba: The only real concern I have here is that this consumer is a single constant used by three controllers (ceilometer, cloudkitty and autoscaling). This means that a ceilometer reconcile strips the finalizer from autoscaling's / cloudkitty's transport secrets and vice-versa.

lib-common's docstring warns about this behaviour: "If multiple CR instances sharing the same consumerFinalizer coexist in one namespace... use a per-instance finalizer."

So this might be a wanted behaviour, I just wanted to highlight it before approving. Everything else LGTM

lmiccini: thanks Juan, I switched to a per-cr finalizer (the stripping race was not the intended behavior).

lmiccini:
recheck

jlarriba (approved):
LGTM

